PR #2, opened 2020-05-28
SANTUARIO-547: Update xml catalog to newer format

lukasj:
Signed-off-by: Lukas Jungmann <lukas.jungmann@oracle.com>


src/main/resources/bindings/bindings.cat
```diff
@@ -1,8 +1,18 @@
...
-SYSTEM "http://www.w3.org/2001/XMLSchema.dtd" "schemas/XMLSchema.dtd"
-
-PUBLIC "datatypes" "schemas/datatypes.dtd"
-SYSTEM "datatypes.dtd" "schemas/datatypes.dtd"
-
-SYSTEM "http://www.w3.org/2001/xml.xsd" "schemas/xml.xsd"
\ No newline at end of file
+<!DOCTYPE catalog
+  PUBLIC "-//OASIS//DTD XML Catalogs V1.1//EN"
+         "http://www.oasis-open.org/committees/entity/release/1.1/catalog.dtd">
```
coheigea, line 3: Can we remove the DocType here, as it introduces a requirement of having an internet dependency?

lukasj: removed

PR #56, opened 2021-08-30
SANTUARIO-576 - SAML Validation fails in WildFly 19, Workaround

dmoroff:
As mentioned in the JIRA issue, this is a workaround to SAML assertions run in an EAR within WildFly 19.


coheigea (commented):
@seanjmullan what do you think of this change? I'm inclined to say that it should be rejected, because "isEqualNode" would return true if there was a different Node in the document with the same name, etc.

seanjmullan:
I agree. We also shouldn't introduce a workaround for an issue where the behavior change is unexplained.

MichaelVetter:
@seanjmullan, @coheigea : What do you think of a similar change that only uses "isEqualNode" if the compared nodes are instances of different classes?

PR #58, opened 2021-09-02
SANTUARIO-577 - Introduce a system property to control if file/http r…

coheigea:
…eferences are allowed from an unsigned context

### DIFF
--- a/src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMURIDereferencer.java
+++ b/src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMURIDereferencer.java
@@ -102,16 +102,19 @@ public final class DOMURIDereferencer implements URIDereferencer {
             }
         }
 
-        try {
-            ResourceResolverContext resContext = new ResourceResolverContext(uriAttr, baseURI, secVal);
-            XMLSignatureInput in = ResourceResolver.resolve(resContext);
-            if (in.isOctetStream()) {
-                return new ApacheOctetStreamData(in);
-            } else {
-                return new ApacheNodeSetData(in);
+        ResourceResolverContext resContext = new ResourceResolverContext(uriAttr, baseURI, secVal);
+        if ((uriRef instanceof javax.xml.crypto.dsig.Reference) || resContext.isURISafeToResolve()) {
+            try {
+                XMLSignatureInput in = ResourceResolver.resolve(resContext);
+                if (in.isOctetStream()) {
+                    return new ApacheOctetStreamData(in);
+                } else {
+                    return new ApacheNodeSetData(in);
+                }
+            } catch (Exception e) {
+                throw new URIReferenceException(e);
             }
-        } catch (Exception e) {
-            throw new URIReferenceException(e);
         }
+        throw new URIReferenceException("URI " + uri + " is forbidden");
     }
 }
--- a/src/main/java/org/apache/xml/security/encryption/XMLCipherInput.java
+++ b/src/main/java/org/apache/xml/security/encryption/XMLCipherInput.java
@@ -115,7 +115,14 @@ public class XMLCipherInput {
             try {
                 ResourceResolverContext resolverContext =
                     new ResourceResolverContext(uriAttr, null, secureValidation);
-                input = ResourceResolver.resolve(resolverContext);
+                if (resolverContext.isURISafeToResolve()) {
+                    input = ResourceResolver.resolve(resolverContext);
+                } else {
+                    String uriToResolve = uriAttr != null ? uriAttr.getValue() : null;
+                    Object[] exArgs = {uriToResolve != null ? uriToResolve : "null", null};
+
+                    throw new ResourceResolverException("utils.resolver.noClass", exArgs, uriToResolve, null);
+                }
             } catch (ResourceResolverException ex) {
                 throw new XMLEncryptionException(ex);
             }
--- a/src/main/java/org/apache/xml/security/utils/resolver/ResourceResolverContext.java
+++ b/src/main/java/org/apache/xml/security/utils/resolver/ResourceResolverContext.java
@@ -18,12 +18,18 @@
  */
 package org.apache.xml.security.utils.resolver;
 
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.Collections;
 import java.util.Map;
 
 import org.w3c.dom.Attr;
 
 public class ResourceResolverContext {
+
+    private static boolean allowUnsafeResourceResolving =
+            AccessController.doPrivileged(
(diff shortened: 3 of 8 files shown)


src/main/java/org/apache/xml/security/utils/resolver/ResourceResolverContext.java
```diff
@@ -51,4 +57,12 @@ public ResourceResolverContext(Attr attr, String baseUri, boolean secureValidati
         return properties;
     }
 
+    public boolean isSafeURIToResolve() {
+        if (allowUnsafeResourceResolving) {
+            return true;
+        }
+        boolean forbiddenURI = (uriToResolve != null && (uriToResolve.startsWith("file:") || uriToResolve.startsWith("http:")))
```
seanjmullan, line 64: Unless this is done somewhere else, I think you should lowercase `uriToResolve` (probably in the ctor) so that it also matches mixed/upper case schemes, ex "HTTP:".

src/main/java/org/apache/xml/security/utils/resolver/ResourceResolverContext.java
```diff
@@ -51,4 +57,12 @@ public ResourceResolverContext(Attr attr, String baseUri, boolean secureValidati
...
+                || (baseUri != null && (baseUri.startsWith("file:") || baseUri.startsWith("http:")));
```
seanjmullan, line 65: For the baseUri, you can make this check more precise by only checking relative URIs. This way you won't trigger false cases where a baseURI has been set for resolving some other URI but not an unsafe reference URI. Something like:

`if (baseUri != null && !baseUri.isEmpty() && baseUri.charAt(0) != '#') {
  // now check if it is file or http ...`

src/main/java/org/apache/xml/security/utils/resolver/ResourceResolverContext.java
```diff
@@ -51,4 +57,12 @@ public ResourceResolverContext(Attr attr, String baseUri, boolean secureValidati
         return properties;
     }
 
+    public boolean isSafeURIToResolve() {
```
seanjmullan, line 60: `isURISafeToResolve` sounds more better to me.

src/main/java/org/apache/xml/security/utils/resolver/ResourceResolverContext.java
coheigea: I think we don't need to do this, as the resource resolvers only handle lower case file/http URLs

src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMURIDereferencer.java
```diff
@@ -102,16 +102,19 @@ public Data dereference(URIReference uriRef, XMLCryptoContext context)
...
+                } else {
+                    return new ApacheNodeSetData(in);
+                }
+            } catch (Exception e) {
+                throw new URIReferenceException(e);
             }
-        } catch (Exception e) {
-            throw new URIReferenceException(e);
         }
+        throw new URIReferenceException("Uri " + uri + " is forbidden");
```
seanjmullan, line 118: Just a nit, change "Uri" to "URI" to be consistent with attribute name.

coheigea: Done

src/main/java/org/apache/xml/security/encryption/XMLCipherInput.java
```diff
@@ -115,7 +115,14 @@ public void setSecureValidation(boolean secureValidation) {
             try {
                 ResourceResolverContext resolverContext =
                     new ResourceResolverContext(uriAttr, null, secureValidation);
-                input = ResourceResolver.resolve(resolverContext);
+                if (resolverContext.isURISafeToResolve()) {
+                    input = ResourceResolver.resolve(resolverContext);
+                } else {
+                    String uriToResolve = uriAttr != null ? uriAttr.getValue() : null;
+                    Object[] exArgs = {uriToResolve != null ? uriToResolve : "null", null};
```
seanjmullan, line 122: I don't think you need to set the first argument to `"null"` if `uriToResolve` is `null`. I think the exception message should be able to handle `null` arguments, i.e. `MessageFormat.format` should print `"null"` for `null` arguments.

src/test/java/org/apache/xml/security/test/dom/utils/resolver/ResourceResolverTest.java
```diff
@@ -96,4 +98,65 @@ public void testLocalFileWithEmptyBaseURI() throws Exception {
...
+        ResourceResolverContext resolverContext =
+                new ResourceResolverContext(uriAttr, file, false);
+        assertFalse(resolverContext.isURISafeToResolve());
+    }
+
+    @org.junit.jupiter.api.Test
+    public void testIsSafeURIToResolveHTTP() throws Exception {
+        Document doc = TestUtils.newDocument();
+        Attr uriAttr = doc.createAttribute("URI");
+        String basedir = System.getProperty("basedir");
```
seanjmullan, line 131: unused variable

src/test/java/org/apache/xml/security/test/dom/utils/resolver/ResourceResolverTest.java
```diff
@@ -96,4 +98,65 @@ public void testLocalFileWithEmptyBaseURI() throws Exception {
...
+        ResourceResolverContext resolverContext =
+                new ResourceResolverContext(uriAttr, null, false);
+        assertFalse(resolverContext.isURISafeToResolve());
+    }
+
+    @org.junit.jupiter.api.Test
+    public void testIsSafeURIToResolveHTTPBaseURI() throws Exception {
+        Document doc = TestUtils.newDocument();
+        Attr uriAttr = doc.createAttribute("URI");
+        String basedir = System.getProperty("basedir");
```
seanjmullan, line 143: unused variable

src/test/java/org/apache/xml/security/test/dom/utils/resolver/ResourceResolverTest.java
```diff
@@ -96,4 +98,65 @@ public void testLocalFileWithEmptyBaseURI() throws Exception {
...
+        ResourceResolverContext resolverContext =
+                new ResourceResolverContext(uriAttr, "http://www.apache.org", false);
+        assertFalse(resolverContext.isURISafeToResolve());
+    }
+
+    @org.junit.jupiter.api.Test
+    public void testIsSafeURIToResolveLocalReference() throws Exception {
+        Document doc = TestUtils.newDocument();
+        Attr uriAttr = doc.createAttribute("URI");
+        String basedir = System.getProperty("basedir");
```
seanjmullan, line 155: unused variable

seanjmullan (approved):

PR #64, opened 2022-04-14
[SANTUARIO-586]sync JDK-8186958: better calculations for HashMap constructor parameter calculation

XenoAmess:
In short, HashMap's int constructor accept capacity, but not size, which means if we wanna not make it resize, it is better to give it capacity calculated using size and default load factor 0.75

See also https://github.com/openjdk/jdk/pull/7928


coheigea (approved):
Any objections to the merge @seanjmullan  ?

seanjmullan:
No. I approve this PR.

seanjmullan (approved):

PR #66, opened 2022-05-18
Add activation exclusion to jakarta.xml.bind-api

jimma:


pom.xml
```diff
@@ -557,6 +558,30 @@
...
+                    <groupId>jakarta.activation</groupId>
+                    <artifactId>jakarta.activation-api</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>jakarta.activation</groupId>
+            <artifactId>jakarta.activation-api</artifactId>
+            <version>${jakarta.activation.api.version}</version>
+            <scope>compile</scope>
```
coheigea, line 584: Why do we need the exclusion and extra dependency, when 2.1.0 is imported already from angus-activation? https://github.com/eclipse-ee4j/angus-activation/blob/92d6c661cd79bcbc69f0cd5ccdf45da4ffdb048a/pom.xml#L36

jimma: @coheigea  For the api dependency, I want to import it directly without a transitive dependency through angus-activation.

coheigea: @jimma Why?

jimma: @coheigea There are 2 reasons :  
1. It's clear to know the project uses api activation 2.1.0
2. We can replace with another alternative implementation in the future.

coheigea (changes requested):
It turns out that Santuario doesn't need activation at all, the tests pass with the exclusion of com.sun.activation, so I think this would be a better change not to introduce more compile time dependencies.

jimma:
@coheigea  You are right ! I removed this new added dependency.  Thanks for review .

coheigea (approved):

PR #67, opened 2022-06-13
Replace cycles with iterator with enhanced-for

turbanoff:
There are few places in code where manual loop is used with Iterator to iterate over Collection.
Instead of manual cycles it's preferred to use enhanced-for cycle instead: it's less verbose, makes code easier to read and it's less error-prone.
It doesn't have any performance impact: java compiler generates similar code when compiling enhanced-for cycle.


coheigea (approved):
Thanks for your contribution!

PR #69, opened 2022-06-28
adding RNG specified via system property

isaacjensen:
Should allow ability to specify a JAVA_OPT for RNG alg.

### DIFF
--- a/src/main/java/org/apache/xml/security/stax/ext/XMLSecurityConstants.java
+++ b/src/main/java/org/apache/xml/security/stax/ext/XMLSecurityConstants.java
@@ -43,12 +43,14 @@ public class XMLSecurityConstants {
     public static final XMLOutputFactory xmlOutputFactoryNonRepairingNs;
 
     private static final SecureRandom SECURE_RANDOM;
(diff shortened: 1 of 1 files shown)


isaacjensen:
We run xmlsec 2.1.6 with this patch locally.

seanjmullan:
For other system properties, we typically prefix them with the package hierarchy to avoid name clashes, so I would recommend:

`org.apache.xml.security.securerandom.algorithm`

Otherwise, this fix looks ok to me, but I think @coheigea should give the approval since this is in the stax implementation.

On a related issue though, I think we should not hard-code "SHA1PRNG" as the default. There are better alternatives in the JDK now. I would replace this with a call to `new SecureRandom()` or possibly `SecureRandom.getInstanceStrong()`. I haven't looked at the usage to see which is more appropriate.

isaacjensen:
Thanks for the comments @seanjmullan. Let's see if we can get @coheigea approval on this.

coheigea (commented):
The 2.1.x-fixes branch isn't active any more, can you submit a PR against master please?

isaacjensen:
@coheigea PR opened against master!

PR #70, opened 2022-07-07
custom algorithms for SecureRandom

isaacjensen:
submitting against master


isaacjensen:
Updated the method header comment that matched your string above. Let me know if this isn't what you meant.

coheigea (approved):

PR #71, opened 2022-08-31
SANTUARIO-593 - Remove here() function support from Apache Santuario

coheigea:


src/test/java/javax/xml/crypto/test/dsig/Baltimore23Test.java
```diff
@@ -219,29 +217,4 @@ public void test_signature_x509_sn() throws Exception {
         assertTrue(coreValidity, "Signature failed core validation");
     }
 
-    @org.junit.jupiter.api.Test
```
seanjmullan, line 222: Can you also add a Baltimore23XalanTest that tests signature.xml using the JSR 105 API with Xalan support like you did for PhaosXMLDSig3Test?

coheigea: Yes, done!

seanjmullan (approved):

PR #90, opened 2022-11-14
Support for EdDSA: Ed25519 and Ed448

jrihtarsic:
The request for upgrade is given in the [SANTUARIO-562](https://issues.apache.org/jira/browse/SANTUARIO-562).
Currently the merge is done for the 2.3.x and it works with the java 8  + bouncy castle.
References are: 

- [RFC 8032] https://www.rfc-editor.org/rfc/rfc8032
- [RFC 9231] https://www.rfc-editor.org/rfc/rfc9231
- [FIPS 186-5]  https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-5-draft.pdf


jrihtarsic:
@coheigea is it possible to provide a feedback on this PR. Is there something else we should do to get the support for EDDSA support in santuario?

coheigea (approved):
LGTM. There is no JSR-105 test added but we can add this. 

@seanjmullan WDYT?

jrihtarsic:
Hi @coheigea 
I have added the tests for EdDSA(Ed22519 and Ed448) XML signature validation. I guess the JSR-105 test means that.

I also add an option to EdDSA-tests to enable  BouncyCastleProvider only when JDK is before the JDK 15.
Because the default java 15+  JCA providers have all the necessary "tools" to create/validate EdDSA signatures, and the BouncyCastleProvider is not needed with java 15+

seanjmullan:
What we are looking for are tests that use the standard Java XML Signature API as defined by JSR 105. You can look at some examples of tests in the src/test/java/javax/xml/crypto/test/dsig directory. You should be able to use them as a template or extend an existing test like ValidateSignatureTest.

coheigea:
Looks good, but could you add a signature creation test as well for JSR-105 API and not just validation?

jrihtarsic:
Hi I have added now also the create signature. If I would add also other EdDSA JSR-105 API specific tests  (as example with the EdDSA private key), I would probably "lock " the code to use JDK 15+. Maybe we can add more tests in a new PR when this feature will be merged to main branch?

coheigea:
I changed the branch from 2.3.x-fixes to main, can you resolve the conflicts please?

jrihtarsic:
I have resolved the conflicts. BTW will be this feature added also to 2.3.x?

coheigea (changes requested):
Yes we can merge to 2.3.x as well if you need it, as it doesn't break backwards compatibility.

jrihtarsic:
Fixed the License headers in a new commit.

jrihtarsic:
> Yes we can merge to 2.3.x as well if you need it, as it doesn't break backwards compatibility.
If that will not cause too much effort for releasing also new 2.3.x version, that would be superb (I can create a new PR for 2.3.x ). Because we are using it in one stable production environment and it would be easier/faster for us to replace library with 2.3.x version.

coheigea (approved):

coheigea:
Don't bother submitting another PR, I will backmerge directly.

PR #101, opened 2023-01-06
Small boost forward - aggregated changes from one stable stage to another

dmatej:
In the beginning was this innocent issue: https://github.com/eclipse-ee4j/glassfish/issues/24138
Then we have found that the XmlSec is to blame. 
Then I have found that I don't like it's state - I had a suspiction about memory leaks, etc.
So I did this ...

### Changes in Artifact

- Updated dependencies
- **Java 11** instead of **Java 8**
  - that's why I changed the major version. 
  - If anyone wants to continue with JDK8 version, he can continue with maintenance in a new branch. But we need bigger progress.
- `System.Logger` instead of **SLF4J**
  - Allows to use SLF4J2/LOG4J2/JUL/GJULE/Any Custom logging framework.
  - No mandatory dependency required
- Added `module-info.java`
  - That's why I tested it in action with the **GlassFish** and **TCK10**
- `AutoCloseable`
  - Just for several types
  - I have found some old discussion about it, but based on obsoleted opinions. Currently not having it was contraproductive and was asking for leaks (I was able to produce some, but not with GF, just in xmlsec unit tests).
- Split `XmlSignatureInput` to several standalone classes based on the type of input data.
  - I still don't think it is alright, but it is still much better than original "god object"
  - `XmlSignatureStreamInput` IS `AutoCloseable`. And still can cause leaks. But changing it would mean much larger changes. At least this type is standalone and specific. Implementation also allows to use custom closeable objects.
- Added some `XmlUtil` methods for simple work with File+Path+Stream, buffers
- Javadocs, formatting

### Changes in Tests

- **PMD** now tests the tests too
- All issues resolved
  - Especially many resource leaks (200?)
- Performance tests moved to maven failsafe plugin and refactored to **JMH**+**JUnit5**
  - Originally excluded
  - Disabled by default, use `-DskipIT=false` to enable them
  - Locally passed, however someone should analyze `decryptDOM`, after it ends there's still heap used, but I couldn't find why.
  - With more changes the performance improved, especially with **bigger buffers** and NIO.
  - Usage of **Shenandoah GC** - it is more aggressive, stats should be real or close to real at least
  - Proven that **StAX** is much faster than **DOM**.
  - Tests results: watch statistics, if the repeating is slower and slower, there's probably some leak (now it probably isn't). Junit tests just basic criteria.
  - Fixed **JaCoCo** configuration.

### Consequences

- It would be good to release at least milestone after some time, so we could use it in MetroWSIT and Glassfish. Required changes are already prepared, it is quite easy and even better, WSIT will not need to repackage xmlsec, GF will not need SLF4J1 in dependencies, everything is more simple and clean.
- Jetty doesn't complain any more in xmlsec tests
- If some company would like to invest to xmlsec refactoring, it would be great ;-)
- TCK tests results are here: https://ci.eclipse.org/jakartaee-tck/job/jakartaee-tck/job/10.0.x/29/testReport/


arjantijms:
@coheigea what do you think?

coheigea:
@dmatej There are way too many changes in this PR. Can you split them out for a single change in a single PR?

dmatej:
This is not a bug fix but rather a refactoring with tested result. How could you do that? Yeah, I could avoid some updates and reordering, but trivial diffs can do gitbot and if I would have to wait for every review for weeks, I would never finish. 

So this is something like a transaction, a change from one consistent state to another.

The last push is just a rebase, I should have done probably rather a merge than this, that's true.

Cutting that to small pieces doesn't bring anything, just more work. Yes, reviewing large pr by reading is not easy, especially with github browser.

This project would need several large refactoring iterations... next generation.

Another option is to fork the project and move quickly forward with own reviews in hours/days, but cooperation is better, even when sometimes somebody comes with scary radical changes (tested!). Another option is to replace this dependency with something in a better shape.

So, simply said, if you refuse this pr, make a decision. But honestly, reviewing it still takes much less time and effort than implementing and testing it. Someone has to do that... both.

arjantijms:
@coheigea technically speaking the PR is already split, as David used many commits instead of one big commit.

Maybe it's convenient for you to review on a commit basis and then (if needed) create small PRs containing just that commit from this one?

coheigea:
@arjantijms Yes a separate PR by commit would be much easier to review if that would be possible?

coheigea:
Alternatively @dmatej , if you can bundle all changes which don't break backwards compatibility into a single PR, I wlll review all those changes at once.

coheigea:
Hello @dmatej @arjantijms , can we try to find a way forward on this please? Forking the project would be a mistake IMO, as someone else would have to maintain the burden of finding and porting security fixes from this project. I'm not asking for a PR per-commit, but just grouping related commits into 4/5 PRs would be workable.

dmatej:
> @coheigea I'm sure 4/5 PRs would be doable indeed. I had just created an internal TODO issue for myself to look at exactly that ;)

If you can do that, I welcome it, otherwise I would have to come to that probably much later.

coheigea:
Any update @arjantijms ?

dmatej:
> Any update @arjantijms ?

It seems I can get there next week, Arjan has too much work on Faces now ;)

src/main/java/org/apache/xml/security/stax/impl/processor/input/AbstractDecryptInputProcessor.java
```diff
@@ -344,7 +343,8 @@
                 XMLStreamReader xmlStreamReader =
                         inputProcessorChain.getSecurityContext().<XMLInputFactory>get(
                                 XMLSecurityConstants.XMLINPUTFACTORY).createXMLStreamReader(
-                                new MultiInputStream(prologInputStream, decryptInputStream, epilogInputStream), StandardCharsets.UTF_8.name());
+                                new MultiInputStream(prologInputStream, decryptInputStream, epilogInputStream),
```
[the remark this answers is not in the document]
dmatej: True, but not changed by this PR at this moment.

src/main/java/org/apache/xml/security/stax/impl/util/DigestOutputStream.java
```diff
@@ -67,9 +66,9 @@
 
     public byte[] getDigestValue() {
         if (isDebugEnabled) {
-            LOG.debug("Pre Digest: ");
-            LOG.debug(stringBuilder.toString());
-            LOG.debug("End pre Digest ");
+            LOG.log(Level.DEBUG, "Pre Digest: ");
+            LOG.log(Level.DEBUG, stringBuilder.toString());
```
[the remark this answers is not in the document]
dmatej: True, but not changed by this PR.

dmatej:
@coheigea I have a question: Are you OK with change of the minimal Java from 8 to 11? Because if not, I can save time, close the PR and fork XmlSec as we need to keep it updated. 

I could rename the PR to "Minimal Java upgraded to 11" too, because many changes depend on that:
1) Module-info
2) System.Logger (the original target was to remove SLF4J from mandatory dependencies)
3) Java9+ deatures like Path.of
4) `--add-exports`
5) Jakarta EE 10 dependencies

It is visible that XmlSec doesn't hurry anywhere and it is close to be really obsoleted. I think it would be better to create at least a prototype of the next XmlSec major version based on this branch and hurry to the year 2023.

dmatej:
> Yep I'm OK with moving to Java 11 as the minimum version for a new 4.0.0 release. I will start going through this PR in more detail shortly

Ok, then I can create yet several standalone PRs to reduce differences.

dmatej:
@coheigea Hi, I tried to reorder commits and separate yet something, but what remained here is mostly SLF4J -> System.Logger, Java11 features, enabled PMD for tests, and refactoring to avoid resource leaks (changed API). Finally I even did not push the reordering, because it could make the commit history less consistent than now.

So this is probably the final state which has to be reviewed, and then we need to choose the right strategy for the release the `4.0.0-M1` at least.

coheigea:
@dmatej Thanks for all your work so far. Looking at the existing PR it would be useful still to extract all the "@Override" changes into a separate PR that could be backported. Or if you prefer I could do this change and you rebase the PR afterwards?

dmatej:
I wanted to override further inflating of the PR, but you are right, it should be done. I will create a new PR for that in few minutes and rebase this after the merge. So this one will be yet bit smaller.

dmatej:
And yet I should transform probably also all those loops. I tried to avoid that, because for i mostly doesn't fail on concurrent changes, but it can still cause errors. With enhanced loops they will become more visible. That might be painful, but we have to go through that once ... I believe there is no such error in xmlsec, so let's try that.

src/main/java/org/apache/xml/security/transforms/implementations/TransformXPath.java
```diff
@@ -88,9 +88,7 @@ protected XMLSignatureInput enginePerformTransform(
             }
             Node xpathnode = xpathElement.getFirstChild();
             if (xpathnode == null) {
-                throw new DOMException(
-                    DOMException.HIERARCHY_REQUEST_ERR, "Text must be in ds:Xpath"
-                );
+                throw new DOMException(DOMException.HIERARCHY_REQUEST_ERR, "Text must be in ds:Xpath");
```
coheigea, line 91: backport

src/main/java/org/apache/xml/security/transforms/implementations/TransformXPath2Filter.java
```diff
@@ -85,10 +85,10 @@ protected XMLSignatureInput enginePerformTransform(
             }
 
             Document inputDoc = null;
```
coheigea, line 87: Backport

src/main/java/org/apache/xml/security/utils/resolver/implementations/ResolverFragment.java
```diff
@@ -83,20 +86,19 @@ public XMLSignatureInput engineResolveURI(ResourceResolverContext context)
...
         result.setExcludeComments(true);
-
         result.setMIMEType("text/xml");
-        if (context.baseUri != null && context.baseUri.length() > 0) {
-            result.setSourceURI(context.baseUri.concat(context.uriToResolve));
-        } else {
+        if (context.baseUri == null || context.baseUri.isEmpty()) {
             result.setSourceURI(context.uriToResolve);
+        } else {
+            result.setSourceURI(context.baseUri.concat(context.uriToResolve));
```
coheigea, line 101: backport

src/main/java/org/apache/xml/security/utils/resolver/implementations/ResolverLocalFilesystem.java
```diff
@@ -74,24 +70,24 @@ public boolean engineCanResolveURI(ResourceResolverContext context) {
...
+        LOG.log(Level.DEBUG, "But I can't");
 
         return false;
     }
 
     private static URI getNewURI(String uri, String baseURI) throws URISyntaxException {
-        URI newUri = null;
-        if (baseURI == null || baseURI.length() == 0) {
+        final URI newUri;
+        if (baseURI == null || baseURI.isEmpty()) {
```
coheigea, line 90: backport

coheigea (changes requested):
I think these are the last changes to be backported in a separate PR if it's not too painful...

dmatej:
Eh, I deleted the branch by mistake (typo in push -> delete branch with typo -> push to the correct branch ... without source branch, so I deleted it instead). Unfortunately pushing it back did not reopen this PR ... so I will recreate the PR after merging those fragments of this PR.

PR #169, opened 2023-05-17
Propagate failures from the NodeFilter via a checked exception

coheigea:
https://issues.apache.org/jira/browse/SANTUARIO-600

### DIFF
--- a/src/test/java/org/apache/xml/security/test/dom/interop/BaltimoreTest.java
+++ b/src/test/java/org/apache/xml/security/test/dom/interop/BaltimoreTest.java
@@ -18,6 +18,7 @@
  */
 package org.apache.xml.security.test.dom.interop;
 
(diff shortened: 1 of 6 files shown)


src/test/java/org/apache/xml/security/test/dom/interop/BaltimoreTest.java
```diff
@@ -277,37 +277,6 @@ public void test_fifteen_external_dsa() throws Exception {
         assertTrue(verify, filename);
     }
 
-    /**
```
seanjmullan, line 280: I think it can be useful to modify and keep this as a negative test, that is make sure it throws an exception when Xalan is not available to process the here() function.

coheigea: Yep fixed now

PR #176, opened 2023-06-12
Cleanups in tests

dmatej:
- Created XmlSecTestEnvironment to manage filesystem dependency
- Fixed many IO leaks in tests - unclosed streams
- Fixed VersionTest
- Moved javax tests to a subpackage; it fixed issues with java modules.

This PR was separated from #101 to minimize differences and should be compilable yet with Java8 too.


dmatej:
Force push: `java.specification.version` returned `1.8` for JDK 1.8, which could not be parsed to `int`, so it returned `null` and escallated by `NPE`. Now it will return `0` just as the original code. The change was made directly in this commit, so I did not create new commit but amended it.

src/test/java/org/apache/xml/security/test/XmlSecTestEnvironment.java
```diff
@@ -0,0 +1,122 @@
...
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.security.GeneralSecurityException;
+import java.security.KeyStore;
+
+/**
```
coheigea, line 29: We don't put @author tags in our code, can it be removed?

coheigea (changes requested):
LGTM, nice improvement to reduce redundant code in the tests. I'll merge once the author tag is removed.

src/test/java/org/apache/xml/security/test/XmlSecTestEnvironment.java
dmatej: Ok, done. I have found yet two other occurences, but I did not touch them:
`xmlsecurity.jin` (some config file, not sure if it is used) and `DomCryptoBinary` (which is copyrighted by Sun)

PR #181, opened 2023-06-21
Using enhanced loops where possible

dmatej:
- Cleanup made by Eclipse IDE and revisited manually

### DIFF
--- a/src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMSignedInfo.java
+++ b/src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMSignedInfo.java
@@ -71,9 +71,9 @@ public final class DOMSignedInfo extends DOMStructure implements SignedInfo {
     private static final String ALGO_ID_MAC_HMAC_NOT_RECOMMENDED_MD5 =
         Constants.MoreAlgorithmsSpecNS + "hmac-md5";
 
-    private List<Reference> references;
-    private CanonicalizationMethod canonicalizationMethod;
-    private SignatureMethod signatureMethod;
+    private final List<Reference> references;
+    private final CanonicalizationMethod canonicalizationMethod;
+    private final SignatureMethod signatureMethod;
     private String id;
     private Document ownerDoc;
     private Element localSiElem;
@@ -100,17 +100,13 @@ public final class DOMSignedInfo extends DOMStructure implements SignedInfo {
         }
         this.canonicalizationMethod = cm;
         this.signatureMethod = sm;
-        this.references = Collections.unmodifiableList(
-            new ArrayList<>(references));
+        this.references = Collections.unmodifiableList(new ArrayList<>(references));
         if (this.references.isEmpty()) {
-            throw new IllegalArgumentException("list of references must " +
-                "contain at least one entry");
+            throw new IllegalArgumentException("list of references must contain at least one entry");
         }
-        for (int i = 0, size = this.references.size(); i < size; i++) {
-            Object obj = this.references.get(i);
+        for (Object obj : this.references) {
             if (!(obj instanceof Reference)) {
-                throw new ClassCastException("list of references contains " +
-                    "an illegal type");
+                throw new ClassCastException("list of references contains an illegal " + obj.getClass());
             }
         }
     }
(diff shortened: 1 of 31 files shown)


src/main/java/org/apache/xml/security/stax/impl/processor/output/AbstractEncryptOutputProcessor.java
```diff
@@ -160,7 +160,7 @@ public void init(OutputProcessorChain outputProcessorChain) throws XMLSecurityEx
             String encryptionSymAlgorithm = securityProperties.getEncryptionSymAlgorithm();
             try {
                 //initialize the cipher
-                String jceAlgorithm = JCEAlgorithmMapper.translateURItoJCEID(encryptionSymAlgorithm);
+                String jceAlgorithm = JCEMapper.translateURItoJCEID(encryptionSymAlgorithm);
```
coheigea, line 163: Was this change intended?

src/main/java/org/apache/xml/security/stax/impl/processor/output/AbstractSignatureOutputProcessor.java
```diff
@@ -181,8 +182,8 @@ protected void setActiveInternalSignatureOutputProcessor(
     protected DigestOutputStream createMessageDigestOutputStream(String digestAlgorithm)
             throws XMLSecurityException {
 
-        String jceName = JCEAlgorithmMapper.translateURItoJCEID(digestAlgorithm);
-        String jceProvider = JCEAlgorithmMapper.getJCEProviderFromURI(digestAlgorithm);
+        String jceName = JCEMapper.translateURItoJCEID(digestAlgorithm);
+        String jceProvider = JCEMapper.getJCEProviderFromURI(digestAlgorithm);
```
coheigea, line 186: Why this change?

src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMSignedInfo.java
```diff
@@ -100,17 +100,13 @@ public DOMSignedInfo(CanonicalizationMethod cm, SignatureMethod sm,
...
-                "contain at least one entry");
+            throw new IllegalArgumentException("list of references must contain at least one entry");
         }
-        for (int i = 0, size = this.references.size(); i < size; i++) {
-            Object obj = this.references.get(i);
-            if (!(obj instanceof Reference)) {
-                throw new ClassCastException("list of references contains " +
-                    "an illegal type");
+        for (Object obj : this.references) {
+            if (!Reference.class.isInstance(obj)) {
```
coheigea, line 108: According to https://rules.sonarsource.com/java/RSPEC-6202, the isInstance operator is preferred when the class name is known at compile time.

coheigea (commented):
Also 2 PMD failures:

[INFO] PMD Failure: org.apache.xml.security.stax.impl.processor.output.AbstractEncryptOutputProcessor:104 Rule:CollapsibleIfStatements Priority:3 These nested if statements could be combined.
[INFO] PMD Failure: org.apache.xml.security.stax.impl.processor.output.AbstractSignatureOutputProcessor:156 Rule:CollapsibleIfStatements Priority:3 These nested if statements could be combined.

src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMSignedInfo.java
dmatej: This is a bit complicated, first I removed the for loop as it should not be possible to have an incompatible element IF user would keep rules (generics). But there is an unit test which proves that it is possible to pass incompatible type. Maybe it should be refactored later - at least I will comply with rules and bring back the `instanceof`.

src/main/java/org/apache/xml/security/stax/impl/processor/output/AbstractSignatureOutputProcessor.java
dmatej: Automatic change made by Eclipse - the method is defined in the parent class, but it is static. There is no such method in `JCEAlgorithMapper`. In fact the `JCEAlgorithmMapper` should NOT inherit from `JCEMapper`, both contain just static methods and both should be final. But that would be a breaking change, however I don't think anyone "extended" these classes.

I would keep it there as it is here, because it is the right direction. In "sanct4" branch we can continue with breaking changes.

coheigea (approved):

PR #186, opened 2023-07-04
Finals

dmatej:
Based on the review of #101 I asked Eclipse IDE to add the `final` keyword to all local variables and all fields where it was possible.


dmatej:
@coheigea Hi, I hope you are not on holidays now ;-)

coheigea:
I was last week but back now :-) IMO these changes are too intrusive and are not necessary. Can we restrict adding final to member variables?

seanjmullan:
I agree with Colm.

dmatej:
> I was last week but back now :-) IMO these changes are too intrusive and are not necessary. Can we restrict adding final to member variables?

Then it is easier for me just to revert all these changes in the #101 . The project doesn't have any rules, so it is just your opinions and I have to respect them. I already spent too much time with this project which was basically just reducing differences between branches.

This PR was to set the same rule everywhere, yet without checkstyle which I would recommend too, BUT ...

PR #192, opened 2023-07-25
New version, Java 11, System.Logger, AutoCloseable, XmlSignatureInput types

dmatej:
This is a replacement for the #101 which I closed when I deleted the branch by my stupid mistake.

In the beginning was this innocent issue: https://github.com/eclipse-ee4j/glassfish/issues/24138
Then we have found that the XmlSec is to blame. 
Then I have found that I don't like it's state - I had a suspiction about memory leaks, etc.
So I did this ...

### Changes in Artifact

- Updated dependencies
- **Java 11** instead of **Java 8**
  - that's why I changed the major version. 
  - If anyone wants to continue with JDK8 version, he can continue with maintenance in a new branch. But we need bigger progress.
- `System.Logger` instead of **SLF4J**
  - Allows to use SLF4J2/LOG4J2/JUL/GJULE/Any Custom logging framework.
  - No mandatory dependency required
- Added `module-info.java`
  - That's why I tested it in action with the **GlassFish** and **TCK10**
- `AutoCloseable`
  - Just for several types
  - I have found some old discussion about it, but based on obsoleted opinions. Currently not having it was contraproductive and was asking for leaks (I was able to produce some, but not with GF, just in xmlsec unit tests).
- Split `XmlSignatureInput` to several standalone classes based on the type of input data.
  - I still don't think it is alright, but it is still much better than original "god object"
  - `XmlSignatureStreamInput` IS `AutoCloseable`. And still can cause leaks. But changing it would mean much larger changes. At least this type is standalone and specific. Implementation also allows to use custom closeable objects.
- Added some `XmlUtil` methods for simple work with File+Path+Stream, buffers

### Changes in Tests

- **PMD** now tests the tests too
- All issues resolved
  - Especially reported resource leaks
- Performance tests moved to maven failsafe plugin and refactored to **JMH**+**JUnit5**
  - Originally excluded
  - Disabled by default, use `-DskipIT=false` to enable them
  - Locally passed, however someone should analyze `decryptDOM`, after it ends there's still heap used, but I couldn't find why.
  - With more changes the performance improved, especially with **bigger buffers** and NIO.
  - Proven that **StAX** is much faster than **DOM**.
  - Tests results: watch statistics, if the repeating is slower and slower, there's probably some leak (now it probably isn't). JUnit tests do test just basic criteria.
  - Fixed **JaCoCo** configuration.

### Consequences

- It would be good to release at least milestone after some time, so we could use it in MetroWSIT and Glassfish. Required changes are already prepared, it is quite easy and even better, WSIT will not need to repackage xmlsec, GF will not need SLF4J1 in dependencies, everything is more simple and clean.
- Jetty doesn't complain any more in xmlsec tests
- If some company would like to invest to xmlsec refactoring, it would be great ;-)


pom.xml
```diff
@@ -95,46 +95,52 @@
...
 
     <dependencies>
         <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-api</artifactId>
-            <version>${slf4j.version}</version>
-            <scope>compile</scope>
+            <groupId>jakarta.activation</groupId>
+            <artifactId>jakarta.activation-api</artifactId>
+            <version>2.1.1</version>
```
coheigea, line 134: Put the version in a variable same as the others

pom.xml
```diff
@@ -442,32 +477,72 @@
...
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
             <plugin>
                  <groupId>org.gaul</groupId>
                  <artifactId>modernizer-maven-plugin</artifactId>
                  <version>2.6.0</version>
                  <configuration>
-                     <javaVersion>${targetJdk}</javaVersion>
```
coheigea, line 470: Couldn't we just keep targetJdk and change to 11?

coheigea (changes requested):

src/main/java/org/apache/xml/security/parser/XMLParser.java
```diff
@@ -27,15 +27,6 @@
...
 
-    /**
-     * Parses a document from the input stream.
-     * Caller is responsible for closing the stream.
-     *
-     * @param inputStream
-     * @param disallowDocTypeDeclarations
-     * @return {@link Document}
-     * @throws XMLParserException
-     */
```
coheigea, line 38: Why removing it?

src/main/java/org/apache/xml/security/signature/XMLSignatureInput.java
```diff
@@ -228,357 +194,278 @@ public Set<Node> getNodeSet(boolean circumvent) throws XMLParserException, IOExc
...
         }
-        if (isOctetStream() || isElement() || isNodeSet()) {
-            Canonicalizer20010315OmitComments c14nizer = new Canonicalizer20010315OmitComments();
-            try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
-                c14nizer.engineCanonicalize(this, baos, secureValidation);
-                bytes = baos.toByteArray();
-            }
+        if (isElement() || isNodeSet()) {
+            ByteArrayOutputStream baos = new ByteArrayOutputStream();
+            canonize(baos, false);
```
coheigea, line 252: Rename to canonicalize

src/main/java/org/apache/xml/security/signature/XMLSignatureInput.java
```diff
@@ -228,357 +194,278 @@ public Set<Node> getNodeSet(boolean circumvent) throws XMLParserException, IOExc
...
-    }
-
-    public void setSecureValidation(boolean secureValidation) {
-        this.secureValidation = secureValidation;
-    }
 
-    public String getPreCalculatedDigest() {
-        return preCalculatedDigest;
+    /**
+     * Canonizes this object to the output stream.
```
coheigea, line 454: Canonicalize

coheigea (changes requested):
LGTM after these changes

src/main/java/org/apache/xml/security/parser/XMLParser.java
dmatej: Mistake - I made a separate PR for master and immediately rebased and reverted unnecessary changes here. Unfortunately I forgot that it is a sequence of commits so the patch was applied by git in the next rebase without conflicts.
Thanks for noticing that!

coheigea (approved):

dmatej:
@coheigea Hi, thank you for all your patience with me! Can you schedule some prerelease, like M1 milestone? I see yet Apache build is failing, because it still depends on Java8, so there is yet some work to be done.

coheigea:
It seems there is some problem with Jenkins and the upgrade, JDK 11 was working fine before. I deleted the JDK 8 build. I'll try to figure out what's going on. We can get a M1 milestone out in a few weeks if it suits then?

dmatej:
> It seems there is some problem with Jenkins and the upgrade, JDK 11 was working fine before. I deleted the JDK 8 build. I'll try to figure out what's going on. We can get a M1 milestone out in a few weeks if it suits then?

The `Java` argument here seems wrong, but I have no idea why is it there:
```
'-Djava.util.logging.config.file=/home/jenkins/jenkins-agent/workspace/Santuario/Santuario' 'Java' 'JDK11/target/test-classes/logging.properties' 
```

Oh, I see, Jenkins doesn't like spaces in build/project names - some paths are escaped then, and some are not.
```
javaagent:/home/jenkins/.m2/repository/org/jacoco/org.jacoco.agent/0.8.10/org.jacoco.agent-0.8.10-runtime.jar=destfile=/home/jenkins/jenkins-agent/workspace/Santuario/Santuario Java JDK11/target/jacoco.exec,includes=org/apache/xml/security/**
```

coheigea:
Thanks for the hint, I renamed the projects to use dashes in the names instead of spaces and it works.

coheigea:
@dmatej Snapshots are available now on https://repository.apache.org/content/groups/snapshots/org/apache/santuario/xmlsec/4.0.0-SNAPSHOT/

arjantijms:
@coheigea would it be an option to also publish an actual non-snapshot, e.g. 4.0.0-M1?

coheigea:
Vote called on 4.0.0-M1 https://lists.apache.org/thread/tbolw0ngfcshyl1wwkc8rdcb9ym1279r

coheigea:
No, I just wanted to check that it wasn't being used any more, thanks!

coheigea:
Here it is: https://repo1.maven.org/maven2/org/apache/santuario/xmlsec/4.0.0-M1/

PR #195, opened 2023-07-31
Make provider in JCEMapper.java Thread-safe

darkmoonka:
-Introduce a ThreadLocal provider configuration that can hold a thread specific default provider. This way concurrency on the default provider can be avoided.

Based on the discussion and points here: https://github.com/apache/santuario-xml-security-java/pull/184#issuecomment-1607276879, I opened this PR to make the default provider configurable per thread. Also this solution is backward compatible, so this should not break any existing integrations.

### DIFF
--- a/src/main/java/org/apache/xml/security/algorithms/JCEMapper.java
+++ b/src/main/java/org/apache/xml/security/algorithms/JCEMapper.java
@@ -37,7 +37,9 @@ public class JCEMapper {
 
     private static Map<String, Algorithm> algorithmsMap = new ConcurrentHashMap<>();
 
-    private static String providerName;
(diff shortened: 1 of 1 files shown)


coheigea:
@seanjmullan Can you review this please?

seanjmullan:
> @seanjmullan Can you review this please?

This fix looks fine to me.

seanjmullan (approved):

darkmoonka:
Thank you guys @coheigea and @seanjmullan !
Do you have some rough ideas when the next release 3.0.3 is to be expected?

coheigea:
@darkmoonka A vote is being called today, expect a release next week.

PR #204, opened 2023-08-28
Adding a "bouncycastle" provider to run the bouncycastle specific tests

coheigea:


src/test/java/org/apache/xml/security/test/dom/algorithms/SignatureAlgorithmTest.java
```diff
@@ -104,18 +104,23 @@ void testConstructionWithProvider() throws Exception {
...
+                            : null;
+                    SignatureAlgorithm signatureAlgorithm = new SignatureAlgorithm(doc, algorithmURI, provider, spec);
+                    assertEquals(provider.getName(), signatureAlgorithm.getJCEProviderName());
+                } catch (XMLSecurityException e) {
+                    assertEquals("", Arrays.asList(e.getStackTrace()).toString());
+                }
             }
+        } catch (ReflectiveOperationException e) {
+            // BouncyCastle not installed, ignore
         }
```
seanjmullan, line 124: It might be useful to add `Assumptions.assumeNotNull(provider)` here so the test is skipped instead of ignored. (You will need to expand the scope of the `provider` object.)

src/test/java/org/apache/xml/security/test/dom/signature/SignatureTest.java
```diff
@@ -93,13 +96,18 @@ void testSigningVerifyingFromExistingSignature() throws Throwable {
...
+            Document doc = getOriginalDocument();
+            XMLSignature signature = signDocument(doc, provider);
+            assertEquals(provider.getName(), signature.getSignedInfo().getSignatureAlgorithm().getJCEProviderName());
+
+            PublicKey pubKey = getPublicKey();
+            assertTrue(signature.checkSignatureValue(pubKey));
+        } catch (ReflectiveOperationException e) {
+            // BouncyCastle not installed, ignore
+        }
     }
```
seanjmullan, line 111: Same comment here as before.

seanjmullan:
Another enhancement related to this that I was thinking is to (optionally) run **all** of the tests with BouncyCastle installed as the first provider. This would give us more confirmation that Santuario works with different JCE providers (those installed in the JDK or BouncyCastle). This might require some Junit wizardry that I'm not an expert enough to know if it is possible or not.

coheigea:
Good point, I added the assumptions. Regarding tests, we could just run the tests twice as part of CI,i.e "mvn clean install; mvn clean install -Pbouncycastle"

seanjmullan:
Thanks. I think you may have misunderstood my latter comment though. What I was thinking if we could run all of the tests with this profile and be sure that BC was used as the default JCE provider for all of the tests, as this would give us a better assurance that all Santuario APIs worked across the board with different JCE providers. Typically this would be done by editing the `java.security` file and adding it as the first provider. For testing, however, modifying the JDK under test is usually not a viable option, so this could be done instead by specifying an alternate `java.security` file with the `java.security.properties` system property, ex:

`java -Djava.security.properties=my.java.security ...`

where `my.java.security` is a file with an entry with the first provider as BC followed by the existing JDK providers, all renumbered, something like:

```
security.provider.1=org.bouncycastle.jce.provider.BouncyCastleProvider
security.provider.2=SUN
security.provider.3=SunRsaSign
security.provider.4=SunEC
security.provider.5=SunJSSE
security.provider.6=SunJCE
security.provider.7=SunJGSS
security.provider.8=SunSASL
security.provider.9=XMLDSig
security.provider.10=SunPCSC
security.provider.11=JdkLDAP
security.provider.12=JdkSASL
security.provider.13=Apple
security.provider.14=SunPKCS11
```

This might be a bit hard to pull off, but would be a very valuable feature as it would provide comprehensive interoperability testing of at least 2 different JCE providers.

PR #218, opened 2023-10-05
SANTUARIO-606 - Avoid creating the SecureRandom in XMLSecurityConstants during class initialization

ppalaga:


src/main/java/org/apache/xml/security/stax/ext/XMLSecurityConstants.java
```diff
@@ -73,17 +74,34 @@ protected XMLSecurityConstants() {
...
+        if (rnd == null) {
+            synchronized (SECURE_RANDOM_LOCK) {
+                rnd = SECURE_RANDOM;
+                if (rnd == null) {
+                    try {
+                        String prngAlgorithm = System.getProperty("org.apache.xml.security.securerandom.algorithm");
+                        SECURE_RANDOM = rnd = prngAlgorithm != null ? SecureRandom.getInstance(prngAlgorithm)
+                                : new SecureRandom();
+                    } catch (NoSuchAlgorithmException e) {
+                        throw new RuntimeException(e);
```
ppalaga, line 96: @coheigea not sure it is worth to cache also the exception. WDYT?

coheigea: I'm not so keen on the synchronized block, would it be possible to use the "Lazy Initialization Holder Class Idiom" (i.e. https://dzone.com/articles/another-singleton-implementation), would this work with Quarkus?

src/main/java/org/apache/xml/security/stax/ext/XMLSecurityConstants.java
```diff
@@ -73,17 +74,34 @@ protected XMLSecurityConstants() {
...
      */
     public static byte[] generateBytes(int length) throws XMLSecurityException {
+
+        SecureRandom rnd = SECURE_RANDOM;
+        if (rnd == null) {
+            synchronized (SECURE_RANDOM_LOCK) {
+                rnd = SECURE_RANDOM;
+                if (rnd == null) {
+                    try {
+                        String prngAlgorithm = System.getProperty("org.apache.xml.security.securerandom.algorithm");
```
coheigea, line 92: Couldn't we keep the System.getProperty as a static variable read at initialization?

coheigea (changes requested):

src/main/java/org/apache/xml/security/stax/ext/XMLSecurityConstants.java
ppalaga: > use the "Lazy Initialization Holder Class Idiom" (i.e. https://dzone.com/articles/another-singleton-implementation), would this work with Quarkus?

Quarkus configures GraalVM native image tool in such a way that all application classes are requested to be initialized at build time (unless configured otherwise for some specific classes). That unfortunately holds also for the inner holder class: its static SecureRandom field would get initialized at build time, unless we configure it to be initialized at runtime (which opens another can of worms - all dependent classes would have to get runtime-initialized too) That would actually mean that our original problem was not solved. 

I am open to suggestions, if you by any chance see some flavor of the pattern where the SecureRandom is not a part of any class initializer.

src/main/java/org/apache/xml/security/stax/ext/XMLSecurityConstants.java
ppalaga: > Couldn't we keep the System.getProperty as a static variable read at initialization?

If we do it, then users of Quarkus native mode would only be able to choose the algorithm at build time. Any change would require a rebuild of the application. I am not sure that would be a good idea.

coheigea (approved):
Good for me, but @seanjmullan can you review it as well please?

src/main/java/org/apache/xml/security/stax/ext/XMLSecurityConstants.java
seanjmullan, line 92: Not sure if this code is intended to run under a Security Manager, but if so, it should be wrapped inside doPrivileged. @coheigea ?

src/main/java/org/apache/xml/security/stax/ext/XMLSecurityConstants.java
```diff
@@ -43,7 +43,8 @@ public class XMLSecurityConstants {
     public static final XMLOutputFactory xmlOutputFactory;
     public static final XMLOutputFactory xmlOutputFactoryNonRepairingNs;
 
-    private static final SecureRandom SECURE_RANDOM;
+    private static SecureRandom SECURE_RANDOM;
```
seanjmullan, line 46: Should this also be `volatile`?

ppalaga: +1, thanks, I thought i made it volatile. Sorry.

src/main/java/org/apache/xml/security/stax/ext/XMLSecurityConstants.java
coheigea: Yes please wrap it in doPrivileged @ppalaga like for https://github.com/apache/santuario-xml-security-java/blob/fa63d6fb5acb0361c491c2c4054ce38eb6907746/src/main/java/org/apache/xml/security/encryption/XMLCipherUtil.java#L37

ppalaga:
Oh, I just noticed, that I have not removed the block initializing `SECURE_RANDOM` from the static initializer. That was supposed to be the main motivation for this PR. Let me fix it quickly.

ppalaga:
e0c409f:

* Addressed the feedback from @seanjmullan 
* Removed the the block initializing `SECURE_RANDOM` from the static initializer
* The system property is read using `AccessController.doPrivileged()`

PR #219, opened 2023-10-06
Update protectAgainstWrappingAttackNodeWrapping for proxy elements

jrihtarsic:
Some specific XML DOM handling implementations are using proxy approach on top of pure DOM implementation, such as: 
com.sun.xml.messaging.saaj:saaj-impl:1.5.3
Where HeaderElement1_2Impl is used on top of the dom structure  element com.sun.org.apache.xerces.internal.dom.ElementNSImpl.

This PR is small update to support also the proxy-element usage for  WrappingAttackNodeWrapping.

### DIFF
--- a/src/test/java/org/apache/xml/security/utils/XMLUtilsTest.java
+++ b/src/test/java/org/apache/xml/security/utils/XMLUtilsTest.java
@@ -1,0 +1,56 @@
+package org.apache.xml.security.utils;
+
+
+import org.junit.jupiter.api.Test;
+import org.w3c.dom.Element;
+
+import javax.xml.namespace.QName;
+import javax.xml.soap.*;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class XMLUtilsTest {
+
+    @Test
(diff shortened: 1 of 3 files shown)


src/test/java/org/apache/xml/security/utils/XMLUtilsTest.java
```diff
@@ -0,0 +1,56 @@
+package org.apache.xml.security.utils;
```
coheigea, line 1: Can you add the standard ASF license header here please as used by other files in the project?

coheigea (changes requested):

src/test/java/org/apache/xml/security/utils/XMLUtilsTest.java
jrihtarsic: Sure, I will do that. 
I have a question: if is it okay to add "third-party library" com.sun.xml.messaging.saaj:saaj-impl to test this (I added that library because we had an issue with it using JDK 11), or is this a bit overkill?

coheigea: Yeah that's fine, it's just a test dependency

PR #223, opened 2023-10-17
Implementation of the ECDH-ES Key AgreementMethod

jrihtarsic:
The Key Agreement method implementation as is described in [XML ENC page](https://www.w3.org/TR/xmlenc-core1/#sec-Alg-KeyAgreement).
The  ECDH-ES is tested with EC key types "secp256r1", "secp384r1", "secp521r1", and X keys:  "x25519", "x448"
Currently, only the ConcatKDF method is supported as a key derivation method. 

The to verify correctness of the implementation the tests cases  from the [w3c test cases from 1-6](https://www.w3.org/2008/xmlsec/Drafts/xmlenc-core-11/test-cases/#sec-KeyAgreement)  are executed on newly added feature.

The feature is implemented for branch 2.3.x with intention to support this feature  for java version 8 (with appropriate provider which support XDH keys ).

The ticket for the feature is here: https://issues.apache.org/jira/browse/SANTUARIO-511


coheigea:
@jrihtarsic Can you rebase this off main instead please and tie it to a JIRA?

coheigea:
Backmerged SANTUARIO-511 to 3.0.x-fixes only, so closing this.

PR #240, opened 2023-10-31
Remove call to Signature.getProvider() in debug log

narras-oss:
The debug log message makes a call to Signature.getProvider() too early.    
This causes Signature.chooseFirstProvider() to be called which matches the first provider always rather than the correct provider based on PrivateKey.getAlgorithm() when there are multiple providers.   
This debug log was changed in this commit and introduced the issue: https://github.com/apache/santuario-xml-security-java/commit/1f4891867a938b29f2f1904be35b07d113d3eda9
Added unit test cases with fake provider

### DIFF
--- a/src/main/java/org/apache/xml/security/algorithms/implementations/SignatureBaseRSA.java
+++ b/src/main/java/org/apache/xml/security/algorithms/implementations/SignatureBaseRSA.java
@@ -61,8 +61,7 @@ public abstract class SignatureBaseRSA extends SignatureAlgorithmSpi {
     public SignatureBaseRSA(Provider provider) throws XMLSignatureException {
         String algorithmID = JCEMapper.translateURItoJCEID(this.engineGetURI());
         this.signatureAlgorithm = getSignature(provider, algorithmID);
-        LOG.log(Level.DEBUG, "Created SignatureRSA using {0} and provider {1}",
-            algorithmID, signatureAlgorithm.getProvider());
+        LOG.log(Level.DEBUG, "Created SignatureRSA using {0}", algorithmID);
     }
 
     Signature getSignature(Provider provider, String algorithmID)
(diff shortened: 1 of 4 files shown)


narras-oss:
@coheigea and @seanjmullan We are unable to upgrade to 3.0.3 or 4.0.0 to get the fix for CVE-2023-44483 because of this issue. Appreciate your attention to this pull request.

seanjmullan:
Sorry for the delay. I will take a look now.

src/main/java/org/apache/xml/security/algorithms/implementations/SignatureBaseRSA.java
```diff
@@ -62,7 +62,7 @@ public SignatureBaseRSA(Provider provider) throws XMLSignatureException {
         String algorithmID = JCEMapper.translateURItoJCEID(this.engineGetURI());
         this.signatureAlgorithm = getSignature(provider, algorithmID);
         LOG.log(Level.DEBUG, "Created SignatureRSA using {0} and provider {1}",
-            algorithmID, signatureAlgorithm.getProvider());
+            algorithmID, provider);
```
seanjmullan, line 65: Good catch, I understand the problem, sorry for not detecting this when I fixed this.

I would just restore the prior code and do not include the provider in the log message, because provider can be null and in that case doesn't really make sense for the log message as I was trying to log the provider that was chosen. So change back to:

`LOG.debug("Created SignatureRSA using {}", algorithmID);`

narras-oss: Thanks, reverted debug line

seanjmullan (approved):

coheigea:
Backmerging to 3.0.x as well.

narras-oss:
@coheigea  Is there an ETA for next release ? We are unable to upgrade to latest version to get the CVE fix until this fix included (other than copy-pasting this class into our application to override it)

coheigea:
@narras-oss What release do you specifically need a fix in?

narras-oss:
This particular pull request (which is merged) is what I am referring to as the fix,  either 3.0.4 or 4.0.1 (next release) would work.

coheigea:
I'm calling a vote on 4.0.1 today with the fix

coheigea:
4.0.1 is in maven central now

ivassile:
@coheigea Do you have ETA when 2.3.5 will be tagged? We have to upgrade to this version to resolve the issue. Thanks!

coheigea:
@ivassile A release vote is now under way on the dev list, please follow that.

PR #292, opened 2024-03-14
Add brainpool curves for EC key generation/cypher 

yklymenko:
The protocol for the communication between energy market participants in Germany enforce to use Brainpool P256r1 elliptic curve for signing and encryption of the soap parts (headers and attachments) .  

This is a part (BDEW-profile) of the bigger AS4 Standart for the communication.  

It would be nice to introduce this curves in santuario-xml-security to be able to use it with wss4j library.


coheigea:
Is there a reason not to also support the twisted curves defined in https://datatracker.ietf.org/doc/html/rfc5639 ?

yklymenko:
> Is there a reason not to also support the twisted curves defined in https://datatracker.ietf.org/doc/html/rfc5639 ?

The reason is simple: Somebody introduced P*r1 curves in org.apache.xml.security.utils.KeyUtils. And I thought, let's do consistent. Actually I need only 256r1 curve

PR #293, opened 2024-03-15
Add brainpool curves for EC key generation/cypher (main branch)

yklymenko:
The protocol for the communication between energy market participants in Germany enforce to use Brainpool P256r1 elliptic curve for signing and encryption of the soap parts (headers and attachments) .

This is a part (BDEW-profile) of the bigger AS4 Standart for the communication.

It would be nice to introduce this curves in santuario-xml-security to be able to use it with wss4j library.

The same as https://github.com/apache/santuario-xml-security-java/pull/292, but for main branch

PR #298, opened 2024-03-22
[SANTUARIO-614] Tests for the EC Brainpool key types

jrihtarsic:
The purpose of the contributions is to verify (and implement unit tests) for signing and encryption with the  EC Brainpool key types. 

For details see the PR  #293 and [SANTUARIO-614](https://issues.apache.org/jira/browse/SANTUARIO-614)

The code is contributed on behalf of the European Commission’s edelivery project to support [eDelivery AS4 2.0 profile](https://ec.europa.eu/digital-building-blocks/sites/pages/viewpage.action?pageId=708412318).


coheigea (commented):
Is it possible to generate the encryption keys dynamically rather than using a keystore?

jrihtarsic:
@coheigea The new tests are carried out using the identical "test pattern" employed for other key types". And these tests uses the certificates, not just the keys .  Is there a particular reason why the keystore approach is not suitable Certificates using this type of key?

jrihtarsic:
@coheigea, I am not sure if this helps, but I updated the instructions for generating certificates, now "the instruction" is using openssl because java/keytool doesn't support these EC key types anymore.

src/test/resources/org/apache/xml/security/samples/input/README.txt
seanjmullan, line 1: I think we should work towards a guideline where we don't accept any more binary files into the source tree, unless there is a very good reason. Binary files are difficult to review, don't fit well with source code repositories and for other reasons, should be discouraged. I have looked at the contents of the `ecbrainpool.p12` file and so I am ok with approving it this time. But for future tests, `keytool` can be used to create certificates and keystores, and supports most common extensions and you can use `ProcessBuilder` to execute it.

seanjmullan (approved):
LGTM. If possible, please fix the typo in the Subject before integrating: s/SANTUIARIO/SANTUARIO/

Also, the subject of SANTUARIO-614 is "edelivery AS4 profile with the brainpool EC keytypes" - I think it should match the title of this PR, which better reflects the change: "Tests for the EC Brainpool key types"

src/test/resources/org/apache/xml/security/samples/input/README.txt
jrihtarsic: I have addressed the ‘title comments’ above. 

Regarding the keystore, I fully agree with the reasoning. Going forward, I will try to use keytool with "ProcessBuilder" to generate certificates before running tests in my future contributions. With the latest enhancements in JDK 17, keytool has become the best tool for this task. 

However, testing certain features and certificates with non keytool supported key types with older JDKs using various JCPs may pose some challenges. But we will tackle that when there :)

coheigea (approved):

PR #299, opened 2024-03-24
[SANTUARIO-615] Implementation pre-post processing extension with xades (basic) example.

jrihtarsic:
The purpose of this PR is to present the proposals [SANTUARIO-615] for the pre/post processor for the signatures.   
The use case is the request for XAdES support as requested in SANTUARIO-402 and SANTUARIO-416. The use-case already supports (Basic Signatures profile) and (with the right certificate) it pass the validation by EC DSS tool: 
https://ec.europa.eu/digital-building-blocks/DSS/webapp-demo/validation 

Please note that for "composing" the "XAdES" data,  the generated jaxb objects are used from the xades scheme v1.4.1 and v1.3.2.
 
 This is a experimental PR to discuss/verify the following architectural approaches for the XAdES:
  - the pre-post processing design pattern for signing (Is the approach accepted, in example is used one interface, or it wold be better to have each own for pre and post processing,..  )
  - use of jaxb generated objects with jakarta.xml.bind.(Marshaller|Unmarshaller).Currently xmlsec is using internal jaxb implementation using ElementProxy for some parts and for others DOMStructure/XMLStructure both with with internal/own/(non generated) "jaxb" object.


coheigea:
@jrihtarsic Sorry for the delay - I think it should be acceptable in theory. It would be good though to split into separate PRs the extensibility changes and then the XAdES implementation.

jrihtarsic:
@coheigea I created now two PRs (and I will close this one)

- [PR 616](https://github.com/apache/santuario-xml-security-java/pull/616)  -  adds the feature to be able to implement pre/post processors to the signature
- [PR 617](https://github.com/apache/santuario-xml-security-java/pull/617) - This PR builds on the feature introduced in PR 616 and provides an implementation of the XAdES‑B profile. In this iteration, the implementation follows the existing Santuario approach by using ElementProxy for marshalling and unmarshalling elements, rather than relying on JAXB.

PR 616 should be merged first. However, if PR 617 is merged before it, then PR 616 can be safely ignored, as PR 617 already includes all of the changes from PR 616.

PR #334, opened 2024-06-18
Fixed KeyValue construction for EC keys

Gamuci:
- The signature KeyValue construction was wrong when an EC key was used. The namespace was declared in the "NamedCurve" element instead of "ECKeyValue". See [https://www.w3.org/TR/xmldsig-core1/#sec-ECKeyValue](https://www.w3.org/TR/xmldsig-core1/#sec-ECKeyValue)
- This fix is needed in Apache Santuario 3.0.x too.


coheigea:
@seanjmullan Please take a look, it looks valid

seanjmullan (approved):

seanjmullan:
> @seanjmullan Please take a look, it looks valid

Yes, I approved.

PR #389, opened 2024-10-07
[SANTUARIO-623] fix the transformation XPATH2 filtering.

jrihtarsic:
This is fix for the issue described in the  [SANTUARIO-623], 
The actual fix is just tin the update of the 

in the XMLSignatureByteInput

* [`src/main/java/org/apache/xml/security/signature/XMLSignatureByteInput.java`](diffhunk://#diff-a5bcc66029146d03000f20be4fff21e48e4751528c3c22f496c03d19d144c14bL70-R74): Refactored the `write` method to improve handling of null `bytes` and streamline output stream checks.
  
Everything else was added just to expand Test Coverage to cover the particular case.


coheigea (approved):

coheigea:
Thanks @jrihtarsic for the excellent work

PR #446, opened 2025-02-21
OID attributes get a wrong name when normalized due to first dot not being removed

baconjander:


coheigea:
@baconjander How did you run into this issue? From the calling code in https://github.com/apache/santuario-xml-security-java/blob/36923f8cced51a0ca38959f1af323a572e98206e/src/main/java/org/apache/xml/security/utils/RFC2253Parser.java#L150 it seems it uses str.substring(i + 1); so I wonder if this is a problem for the normal useage of the library?

baconjander:
Hi,

We run into this problem when a customer configured a certificate for
WSS-Signature which included an OID attribute in the Issuer field. We are
using Axis 1.4, which makes use of WSS4j (and thus xmlsec).

I suppose this case is not very common, the Issuer field normally contains
only the common attributes CN, O, OU, etc. And it is also rare to use
IssuerName + Serial number to identify the Signature certificate, many SOAP
services use BinaryToken instead.

But in that case, the service we ara calling requires IssuerName + Serial
and the Issuer attribute looks like this:

CN=AC Sector Público, OID.2.5.4.97=VATES-Q2826004J, OU=Ceres, O=FNMT-RCM,
C=ES

And after being normalized with RFC2253Parser it becomes like this:

CN=AC Sector Público,.2.5.4.97=VATES-Q2826004J,OU=Ceres,O=FNMT-RCM,C=ES

which doesn't seem to be correct because there is a dot just before the 2
(.2.5.4.97). And the service that receives the SOAP call responds with this
error: "improperly specified input name"

The line 150 that you state is not related with the attribute name
normalization, it is just getting the value of the attribute.

Thanks and best regards


<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/"
xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="
http://www.w3.org/2001/XMLSchema-instance">
<soapenv:Header>
<wsse:Security soapenv:mustUnderstand="1" xmlns:wsse="
http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-secext-1.0.xsd
">
<ds:Signature Id="Signature-1780340652" xmlns:ds="
http://www.w3.org/2000/09/xmldsig#">
<ds:SignedInfo>
<ds:CanonicalizationMethod Algorithm="
http://www.w3.org/2001/10/xml-exc-c14n#"/>
<ds:SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/>
<ds:Reference URI="#id-351288471">
<ds:Transforms>
<ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
</ds:Transforms>
<ds:DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/>
<ds:DigestValue>Ti8r4gwrbe/cO68IdT106pEGHlE=</ds:DigestValue>
</ds:Reference>
</ds:SignedInfo>
<ds:SignatureValue>LMfe0aq5Ge8PKr8gicV18EhZi/sfsl+u0Z1cOWcH3Yen+e6CbCBYFI0T//X26ZE7M744IEhTf70kOjCij9A0vVmMQOHJPN5zZJLAsI+KWDqg+YsVnMBKhhG97hp1RrxL9+yeJrSbvx/CBxgbnSmwfnsbW7pbkcbCPX1X7xdQGdUUrWbzG2iQ96shCMlAyM7PWU26eH9OdBVupO/NKDNfjx7Yoryh9plTtxKy3iyVB93xSs1JaUv3YI7HiZmk21siefRsdtcaPIzXPYFfe/Bey6FhJWGTt0mGwtZyoIQ5QqFSiy9YcMielEtEj62q+O8z2hFVESuTd868Xw2lPdZJww==</ds:SignatureValue>
<ds:KeyInfo Id="KeyId-1754070">
<wsse:SecurityTokenReference wsu:Id="STRId-426816330" xmlns:wsu="
http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-utility-1.0.xsd
">
<ds:X509Data>
<ds:X509IssuerSerial>
<ds:X509IssuerName>CN=AC Sector
P&#xFA;blico,.2.5.4.97=VATES-Q2826004J,OU=Ceres,O=FNMT-RCM,C=ES</ds:X509IssuerName>
<ds:X509SerialNumber>138434699972442398474486497223424998235</ds:X509SerialNumber>
</ds:X509IssuerSerial>
</ds:X509Data>
</wsse:SecurityTokenReference>
</ds:KeyInfo>
</ds:Signature>
<wsu:Timestamp wsu:Id="Timestamp-1181479692" xmlns:wsu="
http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-utility-1.0.xsd
">
<wsu:Created>2025-01-27T15:28:38.833Z</wsu:Created>
<wsu:Expires>2025-01-27T15:33:38.833Z</wsu:Expires>
</wsu:Timestamp>
</wsse:Security>
</soapenv:Header>
<soapenv:Body wsu:Id="id-351288471" xmlns:wsu="
http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-utility-1.0.xsd
">
<RealizarTramiteRequest xmlns="http://representa.seap.minhap.es">
... Ommitted (LOPD :)
</RealizarTramiteRequest>
</soapenv:Body>
</soapenv:Envelope>


Missatge de Colm O hEigeartaigh ***@***.***> del dia dl., 24
de febr. 2025 a les 10:14:

coheigea (approved):

PR #453, opened 2025-02-24
[SANTUARIO-576] Fix saml validation with saaj 1.4+

RivalAUT:
When using SAAJ versions from 1.4 onwards, the signature element of the request was not removed due to type mismatch. This led to failed validations because the digest was wrongly calculated.
With this fix the excludeNode will be excluded even if the currentNode and excludeNode differ in type (e.g. com.sun.xml.messaging.saaj.soap.impl.ElementImpl vs. com.sun.org.apache.xerces.internal.dom.ElementNSImpl).

### DIFF
--- a/src/main/java/org/apache/xml/security/c14n/implementations/CanonicalizerBase.java
+++ b/src/main/java/org/apache/xml/security/c14n/implementations/CanonicalizerBase.java
@@ -241,7 +241,7 @@ public abstract class CanonicalizerBase extends CanonicalizerSpi {
 
             case Node.ELEMENT_NODE :
                 documentLevel = NODE_NOT_BEFORE_OR_AFTER_DOCUMENT_ELEMENT;
-                if (currentNode == excludeNode) {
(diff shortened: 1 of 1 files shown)


RivalAUT:
Sadly not, it does not work in my specific case with just currentNode.isSameNode(excludeNode)
It is needed to check both ways.

coheigea:
It's a bit strange - do you have a test case where isSameNode only works one way?

RivalAUT:
I sent a test case per mail on 13th January.
It happens because currentNode is of type com.sun.org.apache.xerces.internal.dom.ElementNSImpl and excludeNode is of type com.sun.xml.messaging.saaj.soap.impl.ElementImpl, where only checking from com.sun.xml.messaging.saaj.soap.impl.ElementImpl returns the correct result. As both currentNode and excludeNode can be from this class (in my case it is excludeNode, but in the original Jira Ticket it was the other way around), both have to be checked.

seanjmullan (commented):
Look ok, although it may be more optimal to first check if the references are equal and if not then fallback to calling isSameNode. WDYT?

coheigea (approved):

PR #476, opened 2025-05-19
Adding hashcode to DOMKeyValue

coheigea:
A fix for https://github.com/apache/santuario-xml-security-java/security/code-scanning/1143

### DIFF
--- a/src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMKeyValue.java
+++ b/src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMKeyValue.java
@@ -264,6 +264,23 @@ public abstract class DOMKeyValue<K extends PublicKey> extends DOMStructure impl
                 return false;
             }
         }
+
+        @Override
+        public int hashCode() {
+            int result = 17;
+            try {
+                if (getPublicKey() != null) {
+                    RSAPublicKey rsaKey = (RSAPublicKey)getPublicKey();
+                    result = 31 * result + rsaKey.getPublicExponent().hashCode();
+                    result = 31 * result + rsaKey.getModulus().hashCode();
+                }
+            } catch (KeyException ke) {
+                // no key available
+                return super.hashCode();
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMKeyValue.java
```diff
@@ -264,6 +264,22 @@ public boolean equals(Object obj) {
...
+        public int hashCode() {
+            int result = 17;
+            try {
+                if (getPublicKey() != null) {
+                    RSAPublicKey rsaKey = (RSAPublicKey)getPublicKey();
+                    result = 31 * result + rsaKey.getPublicExponent().hashCode();
+                    result = 31 * result + rsaKey.getModulus().hashCode();
+                }
+            } catch (KeyException ke) {
+                // no key available
```
seanjmullan, line 278: I would probably return `super.hashCode(`) here and also change line 264 to return `super.equals()` so if the key can't be extracted only an exact match of the same object will equal.

We should probably add equals/hashCode methods for the other *KeyValue subclasses, though we could do that in another PR. Note that the XMLDSig API does not require implementations to override equals/hashCode. However, it is useful for testing purposes.

src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMKeyValue.java
```diff
@@ -261,9 +261,26 @@ public boolean equals(Object obj) {
                         && rsaKey.getModulus().equals(otherRSAKey.getModulus());
             } catch (KeyException ke) {
                 // no practical way to determine if the keys are equal
-                return false;
+                return super.equals(obj);
```
seanjmullan, line 264: Actually, I take back my comment here - returning false as originally is ok (and better) since it already checked if the objects were the same on line 243.

seanjmullan (approved):

PR #504, opened 2025-08-11
Added properties for base64 formatting

kuzjka:
### New feature

This PR continues [the mailing list thread](https://lists.apache.org/thread/d285pdhjz5mfhcmbrglzkn6qthsyyxjr) and suggests system properties to control base64Binary values formatting:

* `org.apache.xml.security.base64.lineSeparator`, `org.apache.xml.security.base64.lineLength` - allows to override default MIME encoding settings
* `org.apache.xml.security.base64.ignoreLineBreaks` - disables line wrapping for base64 value, but allows to keep the whole XML pretty-printed. Takes precedence over line wrapping settings above.
* existing `org.apache.xml.security.ignoreLineBreaks` takes precedence over all new options.

Default values result to CRLF line breaks with 76 chars in the line, making the whole thing fully compatible with previous implementation.

### Other changes

Some of the logic is included to XMLUtils to provide better encapsulation:
* adding line breaks in the beginning and in the end of wrapped base64 values in the element is implemented as `XMLUtils.encodeElementValue`
* `Base64OutputStream` from Commons, used in XML encryption, is replaced with a `java.util` implementation and provided by `XMLUtils`, making it consistent with configured Base64 encoder.

`Base64.Encoder/Decoder` instances are thread-safe, so sharing a single instance should give us a little performance gain.

### Test coverage

As the base64 configuration takes place during class load, unit-tests leverage from custom `ClassLoader` and Reflection API to reinitialize the class in each test.

### Motivation

https://www.w3.org/TR/xmlschema-2/#base64Binary - `base64Binary` definition has note, that original RFC2045 line-length limitation must not be enforced. In current implementation it is only possible to remove all line breaks in the document using `org.apache.xml.security.ignoreLineBreaks`, sacrificing human readability.

Also, using LF instead of CRLF may be desired in systems where verifying side does not expect escape sequences in base64Binary values.

---

I would be thankful for your comments and suggestions.


src/main/java/org/apache/xml/security/utils/XMLUtils.java
```diff
@@ -1068,4 +1147,52 @@ public static byte[] getBytes(BigInteger big, int bitlen) {
...
+    static class Base64FormattingOptions {
+        private boolean ignoreLineBreaks = false;
+        private Base64LineSeparator lineSeparator = Base64LineSeparator.CRLF;
+        private int lineLength = 76;
+
+        public boolean isIgnoreLineBreaks() {
+            return ignoreLineBreaks;
+        }
+
+        public void setIgnoreLineBreaks(boolean ignoreLineBreaks) {
```
seanjmullan: Instead of set methods, why not add a constructor that takes the options and also checks for illegal syntax?

seanjmullan (commented):
Some comments on a first pass.

src/main/java/org/apache/xml/security/stax/impl/processor/output/AbstractEncryptOutputProcessor.java
```diff
@@ -24,12 +24,7 @@
 import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
 import java.security.spec.AlgorithmParameterSpec;
-import java.util.ArrayDeque;
-import java.util.ArrayList;
-import java.util.Deque;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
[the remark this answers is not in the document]
kuzjka: Oops, my IDE did it :) reverted

src/main/java/org/apache/xml/security/utils/XMLUtils.java
kuzjka: Since we would like to retain meaningful log messages for invalid property values, I couldn't find a better way than to move all properties reading together with validation logic to the constructor of `Base64FormattingOptions`.

Added warnings in case a property is set, but ignored.
Added handling of `NumberFormatException` as suggested in the comment below.

Example log:
```
testIgnoreLineBreaksTakesPrecedence:
... org.apache.xml.security.ignoreLineBreaks property takes precedence over org.apache.xml.security.base64.ignoreLineBreaks, line breaks will be ignored
... Property org.apache.xml.security.base64.lineSeparator has no effect since line breaks are ignored
... Property org.apache.xml.security.base64.lineLength has no effect since line breaks are ignored

testIllegalPropertiesAreIgnored:
... Illegal value of org.apache.xml.security.base64.lineSeparator property is ignored: illegal
... Illegal value of org.apache.xml.security.base64.lineLength property is ignored: illegal
```

src/main/java/org/apache/xml/security/utils/XMLUtils.java
```diff
@@ -1068,4 +1122,90 @@ public static byte[] getBytes(BigInteger big, int bitlen) {
...
+        CRLF(new byte[]{'\r', '\n'}),
+        LF(new byte[]{'\n'});
+
+        private byte[] bytes;
+
+        Base64LineSeparator(byte[] bytes) {
+            this.bytes = bytes;
+        }
+
+        public byte[] getBytes() {
```
seanjmullan: Change to package-private.

src/main/java/org/apache/xml/security/utils/XMLUtils.java
seanjmullan: Did you consider testing the properties by generating signed XML and then checking the format? This test won't catch issues where the internal methods are not called accidentally.

seanjmullan:
> > I think we are good to remove Base64
> 
> Thank you for clearing it up. There are also two duplicating getters `XMLUtils#ignoreLineBreaks()` and `XMLUtils#isIgnoreLineBreaks()`. After removing `Base64` neither of them is used internally. Probably we can deprecate at least one of them.
> 
> Please let me know if you would like me to do this cleanup as a separate change.

Probably best to do this as a separate change just in case we have to later restore them for some reason.

src/test/java/org/apache/xml/security/utils/XMLUtilsTest.java
```diff
@@ -0,0 +1,141 @@
...
+ * <ul>
+ *     <li>{@link XMLUtils#encodeToString(byte[])}</li>
+ *     <li>{@link XMLUtils#encodeElementValue(byte[])}</li>
+ *     <li>{@link XMLUtils#encodeStream(OutputStream)}</li> (creates a wrapper stream, which applies the same encoding
+ *         as {@code encodeToString(byte[])})
+ * </ul>
+ * Output of the first two methods is checked using an appropriate {@link FormattingChecker} implementation.
+ * The result of stream encoding is compared to the output of {@code encodeToString} method.
+ *
+ * There are also tests, which check that the corresponding decoding methods can process Base64-encoded data with any
```
seanjmullan: Nit, no comma needed.

src/test/java/org/apache/xml/security/utils/XMLUtilsTest.java
```diff
@@ -0,0 +1,141 @@
...
+public class XMLUtilsTest {
+
+    private FormattingChecker formattingChecker = FormattingCheckerFactory.getFormattingChecker();
+
+    /* Base64 encoding of the following bytes is: AQIDBAUGBwg= */
+    private static final byte[] TEST_DATA = new byte[]{ 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08 };
+
+    @Test
+    public void testEncodeToString() {
+        byte[] data =  new byte[60]; // long enough for a line break in MIME encoding
```
seanjmullan: nit, extra space after '='. Same comment on other lines.

kuzjka:
Rebased on top of main branch. Please let me know if anything else needs to be done here.

coheigea (approved):
@seanjmullan Are you OK with merging this?

seanjmullan (approved):

seanjmullan:
> @seanjmullan Are you OK with merging this?

Yes, approved.

PR #585, opened 2026-04-13
Refactoring reference resolving

coheigea:


src/test/java/org/apache/xml/security/test/dom/utils/resolver/ResolverLocalFilesystemTest.java
```diff
@@ -0,0 +1,140 @@
...
+        assertFalse(resolver.engineCanResolveURI(ctx),
+            "FIX VERIFIED: http: uriToResolve must be rejected even with a file: baseUri");
+    }
+
+    /**
+     * A relative uriToResolve (no scheme) combined with a file: baseUri is accepted —
+     * this is the primary legitimate use case for providing a file: baseUri.
+     */
+    @Test
+    void testRelativeUriWithFileBaseUriIsAccepted() throws Exception {
```
seanjmullan, line 107: It would be good to add a test with a relative uriToResolve and a non-file baseURI and make sure it is *not* accepted, ex:

`ResourceResolverContext ctx = makeContext("subdoc.xml", "https://victim.com/");`

coheigea: Test added

PR #587, opened 2026-04-14
XMLCipher refactor

coheigea:


src/main/java/org/apache/xml/security/encryption/XMLCipher.java
```diff
@@ -1842,6 +1842,14 @@ public byte[] decryptToByteArray(Element element) throws XMLEncryptionException
         EncryptedData encryptedData = factory.newEncryptedData(element);
         String encMethodAlgorithm = encryptedData.getEncryptionMethod().getAlgorithm();
 
+        // Reject any attempt to decrypt with an algorithm that doesn't match the one specified when the XMLCipher was initialized
+        if (algorithm != null && !algorithm.equals(encMethodAlgorithm)) {
```
seanjmullan: There are several other code paths that call `getEncryptionMethod().getAlgorithm()` - have you checked them for similar issues?

coheigea: @seanjmullan Good catch, there was a similar issue for the key wrap algorithm that I updated in the PR. We use a different approach for streaming so it doesn't apply there.

seanjmullan (approved):

PR #607, opened 2026-06-05
Fixed testAllZerosSignatureRejected - explicit key length

kuzjka:
This makes the test compatible with recent Java versions.

The test relied on default RSA key length, which [was changed](https://github.com/openjdk/jdk/commit/313bc7f64f69d8f352d495d2c35bea62aca910e4#diff-300c011df3fd1eb530bee91a74ce1d00b51502272310a32c6bf673ae298817a1R150) from 2048 to 3072 since jdk-19+16


coheigea:
@kuzjka I fixed this already in a different way, so I think this change isn't required.

kuzjka:
Ok, no problem.
I just noticed the test currently fails on Java 25.
Probably I tested before rebase.

coheigea:
OK I'll close this, if you notice another problem please re-open it.

PR #616, opened 2026-06-15
[SANTUARIO-615] Implements Pre/Post processors for XMLdSig 

jrihtarsic:
This PR adds a new capability to the Santuario library that allows signatures to be processed both before and after generation. The feature is intended to support XAdES functionality (as example), fulfilling the requirements outlined in SANTUARIO‑402 and SANTUARIO‑416.

### DIFF
--- a/src/main/java/module-info.java
+++ b/src/main/java/module-info.java
@@ -32,6 +32,7 @@ module org.apache.santuario.xmlsec {
 
     exports org.apache.jcp.xml.dsig.internal.dom;
     exports org.apache.xml.security;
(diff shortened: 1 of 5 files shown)


jrihtarsic:
Hi @coheigea 
Can you please check this PR. It has the "generic" API updates which allows to add pre-post signature processing which is needed to  implemetne XAdES as example of signature extension.  The XAdES-B is then implemented in [PR 617](https://github.com/apache/santuario-xml-security-java/pull/617) using this API.

coheigea (approved):

PR #627, opened 2026-07-06
PGPData fixes

seanjmullan:
Fix bit masking issue when checking. Check for invalid key packets when unmarshalling PGPData from XML.

### DIFF
--- a/src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMPGPData.java
+++ b/src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMPGPData.java
@@ -157,6 +157,7 @@ public final class DOMPGPData extends DOMStructure implements PGPData {
                 } else if ("PGPKeyPacket".equals(localName) && XMLSignature.XMLNS.equals(namespace)) {
                     String content = XMLUtils.getFullTextChildrenFromNode(childElem);
                     pgpKeyPacket = XMLUtils.decode(content);
+                    checkKeyPacket(pgpKeyPacket);
                 } else {
                     other.add
                     (new javax.xml.crypto.dom.DOMStructure(childElem));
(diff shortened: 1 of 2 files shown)


coheigea (changes requested):
Possible issue:
```
checkKeyPacket throws IllegalArgumentException (unchecked), while the XML constructor is declared [public DOMPGPData(Element pdElem) throws MarshalException]. That mismatch creates an API contract problem: any caller that only catches [MarshalException] to handle all XML-constructor failures will let the IllegalArgumentException escape silently.

Required fix: wrap the call inside the XML constructor to honour its own declared contract:
```

seanjmullan:
> Possible issue:
> 
> ```
> checkKeyPacket throws IllegalArgumentException (unchecked), while the XML constructor is declared [public DOMPGPData(Element pdElem) throws MarshalException]. That mismatch creates an API contract problem: any caller that only catches [MarshalException] to handle all XML-constructor failures will let the IllegalArgumentException escape silently.
> 
> Required fix: wrap the call inside the XML constructor to honour its own declared contract:
> ```

It's not necessary as the exception is later caught and wrapped in a `MarshalException` at the two public entry points for unmarshalling:

- https://github.com/apache/santuario-xml-security-java/blob/42526995105c61ccd604672e400cc7819821b8dc/src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMKeyInfoFactory.java#L183
- https://github.com/apache/santuario-xml-security-java/blob/42526995105c61ccd604672e400cc7819821b8dc/src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMXMLSignatureFactory.java#L227

However, I did improve the test to check that the cause of the `MarshalException` is an `IllegalArgumentException`.

coheigea (approved):

PR #645, opened 2026-08-07
Add PQC (ML-DSA and ML-KEM) support for XML Digital Signature and Enc…

ffang:
…ryption

- ML-DSA (FIPS 204): signature algorithm constants for ML-DSA-44/65/87, JCA mappings and whitelist URIs in XMLSignature, DOM and JSR-105 support.

- ML-KEM (FIPS 203): key transport support using the W3C "XML Security: Generic Hybrid Cipher" structure (https://www.w3.org/TR/xmlsec-generic-hybrid/, closing SANTUARIO-633): KeyUtils.kemEncapsulate/kemDecapsulate perform real javax.crypto.KEM (JEP 452) encapsulation/decapsulation plus HKDF key derivation, with the AES-KeyWrap output and KEM encapsulation concatenated into CipherValue. New ALGO_ID_KEYTRANSPORT_GENERIC_HYBRID top-level algorithm and GenericHybridCipherMethod/KeyEncapsulationMethod/ DataEncapsulationMethod XML structure, wired into both the DOM XMLCipher and STAX (XMLEncryptOutputProcessor/XMLEncryptedKeyInputHandler) paths.

- Both features require BouncyCastle 1.84+ and skip gracefully without it; ML-KEM additionally requires Java 21+ for javax.crypto.KEM, accessed via reflection so the module still compiles under its Java 11 target.

Algorithm URIs follow the RFC 9231 provisional naming pattern under http://www.w3.org/tbd# per draft-eastlake-rfc9231bis-xmlsec-uris, pending final URI assignment.


seanjmullan:
Thank you for this contribution! I plan on reviewing this but will need a couple of weeks.

ffang:
> Thank you for this contribution! I plan on reviewing this but will need a couple of weeks.

Thanks @seanjmullan !

No rush, take your time.

Freeman

seanjmullan:
I like the idea of splitting the PR into 2, one for XML Signature and one for XML Encryption. This will make it easier to focus on one specification at a time. And I think the XML Signature changes should be more straightforward.

ffang:
> I like the idea of splitting the PR into 2, one for XML Signature and one for XML Encryption. This will make it easier to focus on one specification at a time. And I think the XML Signature changes should be more straightforward.

Sure, I will separate the PR and resend.

Cheers
Freeman

Arpan0995:
I have been looking at post-quantum support across a number of Java crypto libraries and read through this PR with interest. Two choices in the KEM path stood out as the right ones: it takes the encapsulation length from the KEM API (Decapsulator.encapsulationSize()) rather than a hardcoded per-algorithm table, and it derives the key-wrap key through a KDF rather than using the raw shared secret.

One gap I noticed while reading the tests. The new PQC tests (XMLSignatureMLDSATest, XMLEncryptionMLKEMTest, StaxMLDSASignatureTest, StaxMLKEMEncryptionTest) all exercise the happy path only, sign-then-verify and encrypt-then-decrypt. For a signature and encryption feature it would be worth adding a few negative cases that lock in the security-relevant behaviour, for example:

- a tampered ML-DSA signature (a flipped byte of SignatureValue) is rejected by verification;
- an ML-DSA signature checked against the wrong public key fails;
- ML-KEM decryption with the wrong recipient private key fails cleanly rather than returning wrong plaintext;
- a truncated or corrupted KEM encapsulation (shorter than encapsulationSize, or a flipped byte) is rejected with a clear exception.

These are the cases that would catch a regression in the verify or decapsulate path later, and they are quick to add on top of the round-trip tests already there. Since you are planning to split this into separate Signature and Encryption PRs, the ML-DSA cases would go with the signature side and the ML-KEM cases with the encryption side. Happy to help with any of them if useful.

ffang:
> I have been looking at post-quantum support across a number of Java crypto libraries and read through this PR with interest. Two choices in the KEM path stood out as the right ones: it takes the encapsulation length from the KEM API (Decapsulator.encapsulationSize()) rather than a hardcoded per-algorithm table, and it derives the key-wrap key through a KDF rather than using the raw shared secret.
> 
> One gap I noticed while reading the tests. The new PQC tests (XMLSignatureMLDSATest, XMLEncryptionMLKEMTest, StaxMLDSASignatureTest, StaxMLKEMEncryptionTest) all exercise the happy path only, sign-then-verify and encrypt-then-decrypt. For a signature and encryption feature it would be worth adding a few negative cases that lock in the security-relevant behaviour, for example:
> 
>     * a tampered ML-DSA signature (a flipped byte of SignatureValue) is rejected by verification;
> 
>     * an ML-DSA signature checked against the wrong public key fails;
> 
>     * ML-KEM decryption with the wrong recipient private key fails cleanly rather than returning wrong plaintext;
> 
>     * a truncated or corrupted KEM encapsulation (shorter than encapsulationSize, or a flipped byte) is rejected with a clear exception.
> 
> 
> These are the cases that would catch a regression in the verify or decapsulate path later, and they are quick to add on top of the round-trip tests already there. Since you are planning to split this into separate Signature and Encryption PRs, the ML-DSA cases would go with the signature side and the ML-KEM cases with the encryption side. Happy to help with any of them if useful.

Hi @Arpan0995 ,

Thanks for the review, and glad the KEM length/KDF choices came through as intended.

Good catch on the negative test coverage — you're right that the current tests only exercise the happy path. I'll send the split PRs (Signature and Encryption) soon and make sure to include the negative cases you listed: tampered signature rejection, wrong-key verification failure, wrong-key decryption failure, and truncated/corrupted encapsulation rejection.

And yes, any help/contribution is welcome — feel free to jump in on either PR once they're up.

Cheers
Freeman

src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMSignatureMethod.java
```diff
@@ -94,6 +94,16 @@ public abstract class DOMSignatureMethod extends AbstractDOMSignatureMethod {
         "http://www.w3.org/2021/04/xmldsig-more#eddsa-ed25519";
     static final String ED448 =
         "http://www.w3.org/2021/04/xmldsig-more#eddsa-ed448";
+
+    // Provisional URIs for ML-DSA (FIPS 204) per draft-eastlake-rfc9231bis-xmlsec-uris
+    // section 3.3.15. These use the draft's "tbd" placeholder namespace and will need
```
bc-pi, line 99: There's been some talk related to https://github.com/w3c/strategy/issues/484 about getting the namespace sorted so maybe coming soon.

ffang: Thanks for the heads-up @bc-pi !

bc-pi: https://www.w3.org/2026/08/xmldsig-more# is looking to be the clubhouse leader

Arpan0995:
Following up on the negative-test cases: since the split PRs are not up yet, I implemented the four cases against the current PQC branch and opened ffang/santuario-xml-security-java#1 into it, so they are ready before the split. All four are parameterized across the three parameter sets in the same style as the existing tests, pass locally with `mvn test -Dtest=XMLSignatureMLDSATest,XMLEncryptionMLKEMTest -P bouncycastle` (9 + 9 executions, 0 failures), and skip via the existing `assumeTrue` guards when BouncyCastle is absent. If you would rather fold them into the split PRs instead, the ML-DSA cases go with the Signature one and the ML-KEM cases with the Encryption one; happy to rebase either way.

ffang:
> Arpan0995

Hi @Arpan0995,

Thanks for going ahead and implementing these — nice catch keeping them parameterized across all three parameter sets, that's cleaner than what I had.

FYI, I'd already been working on adding negative-test coverage, and mine covers both the DOM and STAX paths . But your @ParameterizedTest/@CsvSource approach, parameterized across all three parameter sets, is cleaner than the single hardcoded case I used — I'd like to bring that in.

The split PRs are up now:

- #651 — ML-DSA (signature) — negative-test coverage on both the DOM/JSR-105 and STAX paths (tampered SignatureValue rejected, wrong public key rejected)
- #652 — ML-KEM (encryption) — negative-test coverage on both the DOM and STAX paths (wrong recipient key fails cleanly, truncated EncryptedKey ciphertext rejected)

Would you mind rebasing your parameterization against #651/#652 (ML-DSA cases → #651, ML-KEM cases → #652)? Happy to help however's easiest — a PR against my branch, or a patch pasted here.

Cheers
Freeman

ffang:
Per the discussion we separate this PR with other improvements as suggested here
https://github.com/apache/santuario-xml-security-java/pull/652 (encryption)
https://github.com/apache/santuario-xml-security-java/pull/651 (signature)
So I close this one

Arpan0995:
@ffang Done: the rebased parameterization is up as ffang/santuario-xml-security-java#2 (ML-DSA, into #651) and ffang/santuario-xml-security-java#3 (ML-KEM, into #652).

ffang:
> @ffang Done: the rebased parameterization is up as [ffang#2](https://github.com/ffang/santuario-xml-security-java/pull/2) (ML-DSA, into #651) and [ffang#3](https://github.com/ffang/santuario-xml-security-java/pull/3) (ML-KEM, into #652).

Thanks @Arpan0995 ! Both are merged into my branches

PR #651, opened 2026-08-21
Add ML-DSA (FIPS 204) post-quantum signature support

ffang:
Adds ML-DSA-44/65/87 XML digital signature support via the JSR-105 API (DOM) and the STAX signature path, wired through JCEMapper and the JSR-105 provider's algorithm URI registrations. Part of the post-quantum work tracked under SANTUARIO-634 (originally proposed in SANTUARIO-633 / apache/santuario-xml-security-java#645), split out here as the signature-only half per community request.

- The ML-DSA test keystore is generated on the fly per test run instead of a committed PKCS12 binary, avoiding the maintenance burden of binary test fixtures. Uses SelfSignedCertGenerator, originally authored by Joze Rihtarsic (unmerged PR #617), copied in and extended here with ML-DSA-44/65/87 AlgorithmIdentifier support per his suggestion on #645.
- Adds negative-test coverage on both the DOM/JSR-105 and STAX paths: a tampered SignatureValue is rejected, and verification against the wrong public key fails. Added per Arpan0995's review feedback on #645.

### DIFF
--- a/src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMSignatureMethod.java
+++ b/src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMSignatureMethod.java
@@ -94,6 +94,16 @@ public abstract class DOMSignatureMethod extends AbstractDOMSignatureMethod {
         "http://www.w3.org/2021/04/xmldsig-more#eddsa-ed25519";
     static final String ED448 =
         "http://www.w3.org/2021/04/xmldsig-more#eddsa-ed448";
+
+    // Provisional URIs for ML-DSA (FIPS 204) per draft-eastlake-rfc9231bis-xmlsec-uris
+    // section 3.3.15. These use the draft's "tbd" placeholder namespace and will need
+    // to be updated once final URIs are assigned (see SANTUARIO-634).
+    static final String ML_DSA_44 =
+        "http://www.w3.org/tbd#ml-dsa-44";
+    static final String ML_DSA_65 =
+        "http://www.w3.org/tbd#ml-dsa-65";
+    static final String ML_DSA_87 =
+        "http://www.w3.org/tbd#ml-dsa-87";
     static final String ECDSA_SHA3_224 =
         "http://www.w3.org/2021/04/xmldsig-more#ecdsa-sha3-224";
     static final String ECDSA_SHA3_256 =
@@ -269,6 +279,12 @@ public abstract class DOMSignatureMethod extends AbstractDOMSignatureMethod {
             return new EDDSA_ED25519(smElem);
         } else if (alg.equals(ED448)) {
             return new EDDSA_ED448(smElem);
+        } else if (alg.equals(ML_DSA_44)) {
+            return new MLDSA_44(smElem);
+        } else if (alg.equals(ML_DSA_65)) {
+            return new MLDSA_65(smElem);
+        } else if (alg.equals(ML_DSA_87)) {
+            return new MLDSA_87(smElem);
         } else {
             throw new MarshalException
                 ("unsupported SignatureMethod algorithm: " + alg);
@@ -1289,6 +1305,89 @@ public abstract class DOMSignatureMethod extends AbstractDOMSignatureMethod {
         @Override
         String getJCAAlgorithm() {
             return "Ed448";
+        }
+    }
+
+    abstract static class AbstractMLDSASignatureMethod extends DOMSignatureMethod {
+
+        AbstractMLDSASignatureMethod(AlgorithmParameterSpec params)
+                throws InvalidAlgorithmParameterException {
+            super(params);
+        }
+
+        AbstractMLDSASignatureMethod(Element dmElem) throws MarshalException {
+            super(dmElem);
+        }
+
+        /** ML-DSA signatures are raw bytes; no reformatting needed. */
+        @Override
+        byte[] postSignFormat(Key key, byte[] sig) {
+            return sig;
+        }
+
+        /** ML-DSA signatures are raw bytes; no reformatting needed. */
+        @Override
+        byte[] preVerifyFormat(Key key, byte[] sig) {
+            return sig;
+        }
+
+        @Override
+        Type getAlgorithmType() {
+            return Type.MLDSA;
+        }
+    }
+
+    static final class MLDSA_44 extends AbstractMLDSASignatureMethod {
+        MLDSA_44(AlgorithmParameterSpec params)
+                throws InvalidAlgorithmParameterException {
+            super(params);
+        }
+        MLDSA_44(Element dmElem) throws MarshalException {
+            super(dmElem);
+        }
+        @Override
+        public String getAlgorithm() {
+            return ML_DSA_44;
+        }
+        @Override
+        String getJCAAlgorithm() {
+            return "ML-DSA-44";
+        }
+    }
+
+    static final class MLDSA_65 extends AbstractMLDSASignatureMethod {
+        MLDSA_65(AlgorithmParameterSpec params)
+                throws InvalidAlgorithmParameterException {
+            super(params);
+        }
+        MLDSA_65(Element dmElem) throws MarshalException {
+            super(dmElem);
+        }
+        @Override
+        public String getAlgorithm() {
+            return ML_DSA_65;
+        }
+        @Override
+        String getJCAAlgorithm() {
+            return "ML-DSA-65";
+        }
+    }
+
+    static final class MLDSA_87 extends AbstractMLDSASignatureMethod {
+        MLDSA_87(AlgorithmParameterSpec params)
+                throws InvalidAlgorithmParameterException {
+            super(params);
+        }
+        MLDSA_87(Element dmElem) throws MarshalException {
+            super(dmElem);
+        }
+        @Override
+        public String getAlgorithm() {
+            return ML_DSA_87;
+        }
+        @Override
+        String getJCAAlgorithm() {
+            return "ML-DSA-87";
         }
     }
 }
--- a/src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMSignatureMethod.java
+++ b/src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMSignatureMethod.java
@@ -95,15 +95,14 @@ public abstract class DOMSignatureMethod extends AbstractDOMSignatureMethod {
     static final String ED448 =
         "http://www.w3.org/2021/04/xmldsig-more#eddsa-ed448";
 
-    // Provisional URIs for ML-DSA (FIPS 204) per draft-eastlake-rfc9231bis-xmlsec-uris
-    // section 3.3.15. These use the draft's "tbd" placeholder namespace and will need
-    // to be updated once final URIs are assigned (see SANTUARIO-634).
+    // URIs for ML-DSA (FIPS 204) per draft-eastlake-rfc9231bis-xmlsec-uris-09
+    // section 3.3.15 (see SANTUARIO-634).
     static final String ML_DSA_44 =
-        "http://www.w3.org/tbd#ml-dsa-44";
+        "http://www.w3.org/2026/08/xmldsig-more#ml-dsa-44";
     static final String ML_DSA_65 =
-        "http://www.w3.org/tbd#ml-dsa-65";
+        "http://www.w3.org/2026/08/xmldsig-more#ml-dsa-65";
     static final String ML_DSA_87 =
-        "http://www.w3.org/tbd#ml-dsa-87";
+        "http://www.w3.org/2026/08/xmldsig-more#ml-dsa-87";
     static final String ECDSA_SHA3_224 =
         "http://www.w3.org/2021/04/xmldsig-more#ecdsa-sha3-224";
     static final String ECDSA_SHA3_256 =
--- a/src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMXMLSignature.java
+++ b/src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMXMLSignature.java
@@ -60,11 +60,13 @@ import javax.xml.crypto.dsig.dom.DOMSignContext;
 import javax.xml.crypto.dsig.dom.DOMValidateContext;
 import javax.xml.crypto.dsig.keyinfo.KeyInfo;
 
+import org.apache.xml.security.utils.Constants;
 import org.apache.xml.security.utils.XMLUtils;
 import org.w3c.dom.Attr;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
 
 /**
  * DOM-based implementation of XMLSignature.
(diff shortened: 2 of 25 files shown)


Arpan0995:
While adding inbound-path coverage after the rebase, I found that StAX verification of an ML-DSA signature is not exercised by the current tests, and driving it revealed a round-trip issue with the KeyValue key identifier. Everything below was reproduced on this branch.

**The issue.** With `setSignatureKeyIdentifier(SecurityTokenConstants.KeyIdentifier_KeyValue)`, which is the identifier `StaxMLDSASignatureTest` itself configures, outbound signing with an ML-DSA key succeeds but emits an empty key element:

```xml
<dsig:KeyInfo xmlns:dsig="http://www.w3.org/2000/09/xmldsig#" Id="..."><dsig:KeyValue/></dsig:KeyInfo>
```

Feeding that document back through the library's own inbound side (`InboundXMLSec#processInMessage`) then fails schema validation and never reaches signature verification, even when the signature is valid:

```
XMLSecurityException: cvc-complex-type.2.4.b: The content of element 'dsig:KeyValue' is not complete.
One of '{...DSAKeyValue, ...RSAKeyValue, WC[##other:"http://www.w3.org/2000/09/xmldsig#"]}' is expected.
```

**Root cause.** `XMLSecurityUtils#createKeyValueTokenStructure(AbstractOutputProcessor, OutputProcessorChain, PublicKey)` opens the `dsig:KeyValue` element, then dispatches on `publicKey.getAlgorithm()` with branches for `"RSA"`, `"DSA"` and `"EC"` only. For any other key type it falls through all branches and closes the element, so an ML-DSA key produces `<dsig:KeyValue/>` with no content. (The writer predates this PR and behaves this way for any key type outside those three; this PR is what makes the fall-through reachable, in the very configuration its own test uses.)

**The inbound side of this PR is otherwise sound.** Repeating the same round trip with `KeyIdentifier_NoKeyInfo` and a verification key supplied via `setSignatureVerificationKey`, the inbound path behaves exactly as it should for ML-DSA-65: the valid document verifies, a tampered `SignatureValue` is rejected ("INVALID signature -- core validation failed"), and verification against a wrong public key is rejected the same way. So the gap is confined to the KeyValue serialization, not to the new signature wiring.

**Why the tests do not catch it:** `StaxMLDSASignatureTest` signs through the outbound path but verifies through the DOM engine (`verifyUsingDOM` / `XMLSignature#checkSignatureValue`), so the empty `KeyValue` is never consumed by the inbound processor.

Two possible fixes, in order of preference:

1. Emit `dsig11:DEREncodedKeyValue` (the SubjectPublicKeyInfo encoding) for key types that have no structured KeyValue form. XML Signature 1.1 defines it as a direct `KeyInfo` child, and it would equally be schema-valid inside `dsig:KeyValue` through the same `##other` wildcard this writer already relies on for `dsig11:ECKeyValue`. Carrying an arbitrary public key is exactly what that element exists for, and this PR already adds the ML-DSA names to the DOM `DEREncodedKeyValue` class, so the resolving side has a basis.
2. Failing that, throw at signing time when `KeyIdentifier_KeyValue` is requested for an unsupported key type, rather than emitting XML the library itself will not accept back.

Happy to contribute the inbound-path tests either way (valid round trip, tampered signature, wrong key, all through `processInMessage`, parameterized across the three parameter sets); they would have caught this and would pin the inbound behavior against future refactors. Related note on the encryption split's inbound coverage is on #652.

ffang:
> While adding inbound-path coverage after the rebase, I found that StAX verification of an ML-DSA signature is not exercised by the current tests, and driving it revealed a round-trip issue with the KeyValue key identifier. Everything below was reproduced on this branch.
> 
> **The issue.** With `setSignatureKeyIdentifier(SecurityTokenConstants.KeyIdentifier_KeyValue)`, which is the identifier `StaxMLDSASignatureTest` itself configures, outbound signing with an ML-DSA key succeeds but emits an empty key element:
> 
> ```
> <dsig:KeyInfo xmlns:dsig="http://www.w3.org/2000/09/xmldsig#" Id="..."><dsig:KeyValue/></dsig:KeyInfo>
> ```
> 
> Feeding that document back through the library's own inbound side (`InboundXMLSec#processInMessage`) then fails schema validation and never reaches signature verification, even when the signature is valid:
> 
> ```
> XMLSecurityException: cvc-complex-type.2.4.b: The content of element 'dsig:KeyValue' is not complete.
> One of '{...DSAKeyValue, ...RSAKeyValue, WC[##other:"http://www.w3.org/2000/09/xmldsig#"]}' is expected.
> ```
> 
> **Root cause.** `XMLSecurityUtils#createKeyValueTokenStructure(AbstractOutputProcessor, OutputProcessorChain, PublicKey)` opens the `dsig:KeyValue` element, then dispatches on `publicKey.getAlgorithm()` with branches for `"RSA"`, `"DSA"` and `"EC"` only. For any other key type it falls through all branches and closes the element, so an ML-DSA key produces `<dsig:KeyValue/>` with no content. (The writer predates this PR and behaves this way for any key type outside those three; this PR is what makes the fall-through reachable, in the very configuration its own test uses.)
> 
> **The inbound side of this PR is otherwise sound.** Repeating the same round trip with `KeyIdentifier_NoKeyInfo` and a verification key supplied via `setSignatureVerificationKey`, the inbound path behaves exactly as it should for ML-DSA-65: the valid document verifies, a tampered `SignatureValue` is rejected ("INVALID signature -- core validation failed"), and verification against a wrong public key is rejected the same way. So the gap is confined to the KeyValue serialization, not to the new signature wiring.
> 
> **Why the tests do not catch it:** `StaxMLDSASignatureTest` signs through the outbound path but verifies through the DOM engine (`verifyUsingDOM` / `XMLSignature#checkSignatureValue`), so the empty `KeyValue` is never consumed by the inbound processor.
> 
> Two possible fixes, in order of preference:
> 
>     1. Emit `dsig11:DEREncodedKeyValue` (the SubjectPublicKeyInfo encoding) for key types that have no structured KeyValue form. XML Signature 1.1 defines it as a direct `KeyInfo` child, and it would equally be schema-valid inside `dsig:KeyValue` through the same `##other` wildcard this writer already relies on for `dsig11:ECKeyValue`. Carrying an arbitrary public key is exactly what that element exists for, and this PR already adds the ML-DSA names to the DOM `DEREncodedKeyValue` class, so the resolving side has a basis.
> 
Hi @Arpan0995 ,

Agree that "Emit DEREncodedKeyValue " is better and please go this way.

>     2. Failing that, throw at signing time when `KeyIdentifier_KeyValue` is requested for an unsupported key type, rather than emitting XML the library itself will not accept back.
> 
> 
> Happy to contribute the inbound-path tests either way (valid round trip, tampered signature, wrong key, all through `processInMessage`, parameterized across the three parameter sets); they would have caught this and would pin the inbound behavior against future refactors.
Thanks a lot!
 Related note on the encryption split's inbound coverage is on #652.

Arpan0995:
@ffang Done: the DEREncodedKeyValue fix is up as ffang/santuario-xml-security-java#4 into this branch, with a parameterized test across ML-DSA-44/65/87. As noted there, it makes the KeyValue schema-valid and the embedded key recoverable; full StAX-inbound round-trip additionally needs inbound DEREncodedKeyValue extraction, which I am happy to follow up with separately.

src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMSignatureMethod.java
```diff
@@ -94,6 +94,16 @@ public abstract class DOMSignatureMethod extends AbstractDOMSignatureMethod {
         "http://www.w3.org/2021/04/xmldsig-more#eddsa-ed25519";
     static final String ED448 =
         "http://www.w3.org/2021/04/xmldsig-more#eddsa-ed448";
+
+    // Provisional URIs for ML-DSA (FIPS 204) per draft-eastlake-rfc9231bis-xmlsec-uris
+    // section 3.3.15. These use the draft's "tbd" placeholder namespace and will need
+    // to be updated once final URIs are assigned (see SANTUARIO-634).
```
seanjmullan, line 100: Draft 9 is now using `xmldsig-more` instead of `tbd` so you may as well go ahead and use that, as that will probably be the final URI.

src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMSignatureMethod.java
```diff
@@ -269,6 +279,12 @@ static SignatureMethod unmarshal(Element smElem) throws MarshalException {
             return new EDDSA_ED25519(smElem);
         } else if (alg.equals(ED448)) {
             return new EDDSA_ED448(smElem);
+        } else if (alg.equals(ML_DSA_44)) {
```
seanjmullan, line 282: Something to think about as a (perhaps subsequent) improvement - we could store all the algorithm `String`s in a static `Map` (then you could just do a `Map.containsKey()`) and reduce the number of classes by passing in arguments (stored with each map entry). Most of these subclasses are the same except for algorithm names. There are a few special cases though for algs that take parameters but I think the code could be structured differently for those to create them by the caller and only if needed.

src/main/java/org/apache/xml/security/algorithms/implementations/SignatureMLDSA.java
```diff
@@ -0,0 +1,208 @@
...
+
+    public SignatureMLDSA(Provider provider) throws XMLSignatureException {
+        String algorithmID = JCEMapper.translateURItoJCEID(this.engineGetURI());
+        LOG.log(Level.DEBUG, "Created SignatureMLDSA using {0}", algorithmID);
+
+        try {
+            if (provider == null) {
+                String providerId = JCEMapper.getProviderId();
+                if (providerId == null) {
+                    this.signatureAlgorithm = Signature.getInstance(algorithmID);
```
seanjmullan, line 62: I think you can dismiss all these alerts.

src/main/java/org/apache/xml/security/algorithms/implementations/SignatureMLDSA.java
```diff
@@ -0,0 +1,208 @@
...
+import org.apache.xml.security.signature.XMLSignature;
+import org.apache.xml.security.signature.XMLSignatureException;
+import org.apache.xml.security.utils.XMLUtils;
+
+/**
+ * ML-DSA (FIPS 204) signature algorithm implementation for XML-Dsig.
+ * Supports ML-DSA-44 (NIST security level 2), ML-DSA-65 (level 3),
+ * and ML-DSA-87 (level 5). Requires BouncyCastle 1.81+ as the JCA provider.
+ */
+public abstract class SignatureMLDSA extends SignatureAlgorithmSpi {
```
seanjmullan, line 44: draft-eastlake-rfc9231bis (3.3.15) allows an optional `SignatureContext` element to be specified as a parameter - however the `java.security.Signature` APIs do not support that parameter (see the Non-Goals of JEP 497 for more details). I think you should check for that parameter and if specified, throw an Exception that it isn't supported (i.e. bail out and do not verify or sign the signature).

src/test/java/org/apache/xml/security/test/javax/xml/crypto/dsig/XMLSignatureMLDSATest.java
```diff
@@ -0,0 +1,214 @@
...
+    private static boolean mlDsaAvailable;
+    private static boolean bcAddedForTheTest;
+    private static KeyStore keyStore;
+
+    @BeforeAll
+    static void setUp() {
+        Security.insertProviderAt(
+                new org.apache.jcp.xml.dsig.internal.dom.XMLDSigRI(), 1);
+
+        if (Security.getProvider("BC") == null) {
```
seanjmullan, line 84: The JDK supports ML-DSA in JDK 24 and later and it will also be backported to many earlier releases. It would be good to test against the JDK's implementation of ML-DSA as well as BC. Can we make the choice of provider configurable and based on what release you are running on? Suggest for now only falling back to the class name if `Security.getProvider("BC") == null` AND JDK is less than 24.

ffang: Hi @seanjmullan ,

Thanks for the feedback!

I will revise accordingly.
Freeman

src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMSignatureMethod.java
ffang: resolved by this commit
https://github.com/apache/santuario-xml-security-java/pull/651/commits/a25d9282fc87697a5d54238dbaada726183a3f1c

src/main/java/org/apache/xml/security/algorithms/implementations/SignatureMLDSA.java
ffang: resolved by this commit
https://github.com/apache/santuario-xml-security-java/pull/651/commits/0189ba1ffcda26dff6a7223b07c80aaea55b27c7

src/test/java/org/apache/xml/security/test/javax/xml/crypto/dsig/XMLSignatureMLDSATest.java
ffang: resolved by this commit
https://github.com/apache/santuario-xml-security-java/pull/651/commits/0189ba1ffcda26dff6a7223b07c80aaea55b27c7

src/main/java/org/apache/xml/security/keys/content/DEREncodedKeyValue.java
```diff
@@ -120,8 +121,11 @@ public PublicKey getPublicKey() throws XMLSecurityException {
                 if (publicKey != null) {
                     return publicKey;
                 }
-            } catch (NoSuchAlgorithmException | InvalidKeySpecException e) { //NOPMD
-                // Do nothing, try the next type
+            } catch (NoSuchAlgorithmException | InvalidKeySpecException | RuntimeException e) { //NOPMD
```
seanjmullan, line 124: Suggest logging the exception; otherwise it is very difficult to determine what the problem is.

ffang: Thanks @seanjmullan for the feedback!
Addressed accordingly by commit here https://github.com/apache/santuario-xml-security-java/pull/651/commits/e6960fb3c5a6f82669b8b96157864bf79eff256f

src/main/java/org/apache/jcp/xml/dsig/internal/dom/DOMXMLSignature.java
```diff
@@ -339,6 +343,30 @@ public boolean validate(XMLValidateContext vc)
...
+     * (draft-eastlake-rfc9231bis-xmlsec-uris-09, section 3.3.15). The
+     * {@code java.security.Signature} API offers no way to pass a signature context
+     * to ML-DSA (see the Non-Goals of JEP 497), so such a signature can be neither
+     * created nor verified correctly here; bail out rather than silently ignoring
+     * the context.
+     */
+    private static void checkForUnsupportedSignatureContext(Element sigElem)
+        throws XMLSignatureException
+    {
+        if (sigElem == null) {
```
seanjmullan: You could also return if the algorithm is not ML-DSA.

